PR #6, opened 2020-07-14
fix main diagram url

ferantivero:


ckittel:
Looks like I'll have to adjust the branch policy here to make sure we can all merge (without official review).

PR #34, opened 2020-08-04
Remove --admin -- not suggested to use with AAD group bindings in place.

ckittel:
* Switch the order of the parameters to make it easy for folks who are not using the variable
* no need for `--admin` -- that's the point of our AAD integration


ckittel:
Maybe you're saying the same thing as me here, but to be clear it has nothing to do with the user that is deploying the cluster, it's got everything to do with the user that they logged in with when they execute the `az aks get-credentials` command.  That pops the AAD login process for a device login flow.  As long as you log in as a user in the group we told them to set up, they are good.  We really don't want people to use the `--admin` command as misses the point of the roles we assigned.

ckittel:
@ferantivero - just updated with more context around this, would appreciate another review.

PR #38, opened 2020-08-05
Add Contoso Bicycle background info here

ckittel:
The dedicated Contoso Bicycle page is being removed from the main article, and since we reference it, I'm salvaging it and bringing it here into the repo.  Only minor tweaks to it so that it feels like it's "here" instead of "there."


contoso-bicycle/README.md
```diff
@@ -0,0 +1,115 @@
...
+
+### Networking team
+
+Make sure that different architectural components can talk to each other in a secure manner. They manage the hub and spoke network topologies and IP space allocation.
+
+### Operations team
+
+Responsible for the infrastructure deployment and day-to-day operations of the Azure environment.
+
+## Business requirements
```
ferantivero, line 45: aren't we already covering this whole topic from the docs? Could we just cross-link or we do really want this in here.

contoso-bicycle/README.md
```diff
@@ -0,0 +1,115 @@
...
+- Automated deployments: They understand the importance of automation. They build automated processes for all infrastructure so that environments and workloads
+can easily be recreated consistently and at any time.
+
+### Cost optimization
+
+- Cost center: There’s only one line-of-business. So, all costs are billed to a single cost center.
+
+- Budget and alerts: They have certain planned budgets. They want to be alerted when certain thresholds like 50%, 75%, and 90% of the plan has been reached.
+
+## Design and technology choices
```
ferantivero, line 85: have we considered moving this to main README where we list all the components/technologies in this RI?

contoso-bicycle/README.md
ckittel: Not as sucinctly -- this list is scattered to the wind now.

contoso-bicycle/README.md
ckittel: I had not considered that.  I was trying to salvage this before the next docs publish (in about two hours), didn't try to merge it in.  Just took mostly "as is" for now.

ckittel:
Had a convo on the side with @ferantivero -- we'll merge in as-is for now and consider coming back and integrating this in a different way.

PR #51, opened 2020-08-12
add a remark.js presentation based on the content of this repo.

dcasati:
this is a presentation based on the content of the repo making it simpler for the field to showcase the AKS Secure Baseline pattern when presenting for Partners and Customers.


ckittel:
Thanks!

PR #55, opened 2020-08-18
add a -o tsv flag to the K8S_RBAC_AAD_PROFILE_TENANT_DOMAIN_NAME

dcasati:
force `-o tsv` to the output of K8S_RBAC_AAD_PROFILE_TENANT_DOMAIN_NAME otherwise the sed regex will fail for anyone with a different output format (e.g.: table).


ckittel:
caught this in the prior merged PR, will just close this one.  Thanks :)

PR #63, opened 2020-08-31
setup GitHub CD workflow

faister:
review workflow for CD of AKS cluster


faister:
looking good, go ahead

Issue #70, opened 2020-09-02
Consider updating location default value to use dynamic expression


neilpeterson:
The location parameter default value is set as `eastus2`. Consider updating with the expression `[resourcegroup().location]`.

https://github.com/mspnp/aks-secure-baseline/blob/main/cluster-stamp.json#L43

Understood that the allowed values are limited to availability set enable regions, I can confirm that the ARM validation API will evaluate the expression and enforce the allowed values.

![image](https://user-images.githubusercontent.com/7844635/91932083-0f0f4880-ec9a-11ea-9329-b9d692259663.png)

neilpeterson:
Some additional considerations:

https://docs.microsoft.com/en-us/azure/azure-resource-manager/templates/template-best-practices#location-recommendations-for-parameters

PR #71, opened 2020-09-02
Updated location default value == resource group location

neilpeterson:
Fixes https://github.com/mspnp/aks-secure-baseline/issues/70

### DIFF
--- a/cluster-stamp.json
+++ b/cluster-stamp.json
@@ -40,7 +40,7 @@
             }
         },
         "location": {
-            "defaultValue": "eastus2",
+            "defaultValue": "[resourceGroup().location]",
             "type": "string",
             "allowedValues": [
                 "centralus",
--- a/networking/hub-default.json
+++ b/networking/hub-default.json
@@ -3,6 +3,7 @@
     "contentVersion": "1.0.0.0",
     "parameters": {
         "location": {
+            "defaultValue": "[resourceGroup().location]",
             "type": "string",
             "allowedValues": [
                 "centralus",
--- a/networking/hub-default.json
+++ b/networking/hub-default.json
@@ -3,7 +3,6 @@
     "contentVersion": "1.0.0.0",
     "parameters": {
         "location": {
-            "defaultValue": "[resourceGroup().location]",
             "type": "string",
             "allowedValues": [
                 "centralus",
--- a/networking/hub-regionA.json
+++ b/networking/hub-regionA.json
@@ -9,7 +9,7 @@
             }
         },
         "location": {
-            "defaultValue": "[resourceGroup().location]",
+            "defaultValue": "eastus2",
             "type": "string",
             "allowedValues": [
                 "centralus",
--- a/networking/hub-regionA.json
+++ b/networking/hub-regionA.json
@@ -9,7 +9,7 @@
             }
         },
         "location": {
-            "defaultValue": "eastus2",
+            "defaultValue": "[resourceGroup().location]",
             "type": "string",
             "allowedValues": [
                 "centralus",
--- a/networking/spoke-BU0001A0008.json
+++ b/networking/spoke-BU0001A0008.json
@@ -9,7 +9,6 @@
             }
         },
         "location": {
-            "defaultValue": "[resourceGroup().location]",
             "type": "string",
             "allowedValues": [
                 "centralus",
--- a/networking/spoke-BU0001A0008.json
+++ b/networking/spoke-BU0001A0008.json
@@ -9,6 +9,7 @@
             }
         },
         "location": {
+            "defaultValue": "[resourceGroup().location]",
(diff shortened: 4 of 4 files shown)


neilpeterson:
I have been working through the template aligning a few things with the 'best practices' described in the quick start repo. The proposed changes follow the recommendations found here:

https://github.com/Azure/azure-quickstart-templates/blob/master/1-CONTRIBUTION-GUIDE/best-practices.md#parameters

I agree that the expression looks funny as presented during a portal deployment, but I do like the dynamic nature and potential to simplify the deployment command.

I have only been focused on the cluster deployment. If agreed, I can update the other templates to align and also ensure the docs gel with the changes.

hallihan:
I like the idea of making all the location parameters defaultValues `[resourceGroup().location]` to be consistent with the best practices from azure-quickstart-templates.

ckittel:
Yeah, the only trick is the hub-spoke.  The RG for hub is supposed to hold all of an org's regional hubs.  So the default for the hub resource group would not be the right region most of the time, and regional hubs are such intentional artifacts, they are rarely "left up to chance" to decide what region they are for.  Also, we have one RG to handle all spokes (another common enterprise pattern), which means those spokes will span regions.  I'm a little hesitant to default there because we might be misrepresenting the intentionality of those choices.  the cluster... for sure, no reason not to default that one.

cluster-stamp.json
```diff
@@ -40,7 +40,7 @@
             }
         },
         "location": {
-            "defaultValue": "eastus2",
+            "defaultValue": "[resourceGroup().location]",
```
ckittel, line 43: Forgot one thing.  One reason we hardcoded this was to ensure we showed the intential parity between this location and our paired region in `geoRedundancyLocation` -- if we change this to have a "by convention" here for the default Value, the paired hard-coded geoRedundancyLocation would no longer make sense for all but one region.  Maybe this PR could make `geoRedundancyLocation` a param without a default value and force the user to set that one?

ckittel (changes requested):

ferantivero:
Please let me bring some of my own learnings (understanding) while conducting similarly or same discussions in the past 😸

The `default-value` + `allowed-values` (latter also a DO-NOT by the best practices, with a very special exception it seems might be running into) are - in this very special case - a list of pre-approved locations that had been carefully selected and tested by the Contoso Bicycle infrastructure team to be sure that all the org-allowed/required Azure services are supported. 

This sort of stamped ARMs (fixed) combined with their (dev/test/prod) param files are the kind of aspect we were envisioning back then for real world ARMs. It means, different instances of them with different names like `cluster-stamp-BU0001A0008.json` + `cluster-stamp-BU0001A0008.paramaters.prod.json`, pushed to your very own VCS taking into account that this is what is going to trigger a new cluster deployment.

By the way, originally it wasn't even a `default-value` in there IIRC, somehow similar to what @hallihan mentions at the beginning of this thread.

Later we agreed based on similar discussions to have a default one, which I'd agree does not seem to the best approach today.

neilpeterson:
There is allot of background information here, I am trying to conceptualize all of it, however do apologize I don't think it's all going to click for me.

@ckittel I understand the intention of these deployment are to serve as an example / learning tool. I wonder if we need to spend some more time considering reusability vs. learning and where to place emphasis on each. For example, would it make sense to optimize the tempalte for max re-use and then use documentation and command line parameter input for the example / learning aspect?

I've gone ahead an added updates for all location parameters. My intention is to make the deployments a bit more generalized / flexible for max re-use.  We can take these, or punt while discussing.

ckittel:
Let's not do the regional hub ones unless we change the words to go along with them.  Otherwise we have a mis-match in intent.  We okay to just do the cluster stamp, as that wouldn't involve any narative changes.  If we do the regional ones, we'll probably want to update the words around the deployment.  In an actual org, the "RegionA" in the title would likely be more like the actual region name and it wouldn't even be a parameter.  I agree the balance between making someone feel "at home" in these templates vs "generic reuse" which isn't what an org would be used to seeing, will be tricky.  If we can make the change just to the cluster stamp, we can address the reuse (which our customers don't care about) vs "familiarity" (which will make customers feel more at home) in the upcoming convos about these?

neilpeterson:
Thanks @ckittel, I've adjusted per your message.

cluster-stamp.json
ckittel: @neilpeterson -- wanted to make sure you saw this still in the chatter. What would be a good solution for this?

neilpeterson: Thanks Chad. I think I am tracking some of the intricacies here, but perhaps we should talk it out.

neilpeterson:
It seems there is allot to consider here. I am going to go ahead and close and will re-open once sorted.

PR #74, opened 2020-09-04
put PDB into the correct namespace and use full label

### LINKED ISSUE #47: Azure Advisor: Still Receiveing PDB Warnings
We had done work to make sure we were not getting any Pod Disruption Budget warnings out of Azure Advisor, but we are still getting them on this.  Maybe it's out of our control?  If so, we could considering list that on one of the final pages of the readme here so customers are not surprised, maybe where we talk about reviewing Azure Advisor Alerts.

Maybe better, we can see if we can resolve the issue so we don't get the AA alert.

ckittel:
PDB wasn't being applied to the aspnetapp workload as expected, wasn't in the right namespace and the label didn't match anything.

Fixes #47

### DIFF
--- a/workload/aspnetapp.yaml
+++ b/workload/aspnetapp.yaml
@@ -75,11 +75,12 @@ apiVersion: policy/v1beta1
 kind: PodDisruptionBudget
 metadata:
   name: aspnetapp-pdb
+  namespace: a0008
 spec:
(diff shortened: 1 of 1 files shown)


ferantivero (approved):
LGTM :shipit:

PR #77, opened 2020-09-09
Remove invalid github workflow replacement

ferantivero:
remove invalid replacement for `resource-group-localtion`


ckittel (approved):
Good catch on the dup (and it's typo)

PR #85, opened 2020-09-21
Adding in export command to variable assignments

raykao:
Making this consistent to reduce confusion for those new to bash why some variable assignments included the export command and others did not.


ckittel (approved):
Good catch, thanks!

Issue #69, opened 2020-09-02
set -e does not work in inner-loop-scripts


ccyflai:
There are several "kubectl --watch" command in the script 1-cluster-stamp.sh and "set -e" will break the script with Ctrl-C.

ckittel:
Thanks for the feedback.  @dcasati -- since that change came from you, do you think you might be able to come up with a good compromise here?

dcasati:
Hi @ccyflai  - 

thanks for reporting that.  since we can't trap INT when `errexit` is on. I think think that you want to bail out when one of the previous commands didn't complete. A better way would be to refactor some of the commands into functions but more on that later.

For now, I'd propose removing the set of `errexit`.

```
diff --git a/inner-loop-scripts/shell/1-cluster-stamp.sh b/inner-loop-scripts/shell/1-cluster-stamp.sh
index 48d61e7..41b21ad 100755
--- a/inner-loop-scripts/shell/1-cluster-stamp.sh
+++ b/inner-loop-scripts/shell/1-cluster-stamp.sh
@@ -98,6 +98,8 @@ az ad group member add --group k8s-view-clusterrole --member-id $AKS_ENDUSR_OBJE
 az login
 az account set -s  $MAIN_SUBSCRIPTION
 
+# unset errexit as per https://github.com/mspnp/aks-secure-baseline/issues/69
+set +e
 echo $'Ensure Flux has created the following namespace and then press Ctrl-C'
 kubectl get ns a0008 --watch
```

comments? ok?
-dcasati

PR #86, opened 2020-09-22
fix errexit as per https://github.com/mspnp/aks-secure-baseline/issues/69

dcasati:
fix errexit as per https://github.com/mspnp/aks-secure-baseline/issues/69


ckittel (approved):
🚀

PR #87, opened 2020-09-23
Remove Azure Policy Preview Step (GA now)

ckittel:


hallihan:
@ckittel  Should we hold off on this until after 9/29?  

> Important :
> The general availability (GA) of Azure Policy on AKS is actively releasing across all regions. The expected global completion of the GA release is 9/29/2020. Usage in regions without the GA release requires preview registration steps. However, this will be automatically updated to the GA release when available in the region.

ckittel:
That is correct, that it's rolling out, hence me saying no rush on this.  However, the instructions are no longer clear because this link takes them to a place that doesn't have instructions clearly included anymore.  MOST of the regions we have in the region allow-list for this content already likely already have the feature.  I think, for most consumers of this content, removing the instruction will be more accurate -- and if someone opens an issue around this, we can quickly drop in the registration instructions OR tell them to try again after 29-Sep.

hallihan (approved):
looks good, :shipit:

ckittel:
Thanks for the review and the check on "should we do this now or wait" -- glad you asked!

PR #101, opened 2020-10-29
Add Ingress Controller to ACR

### LINKED ISSUE #98: Remove dependencies on public container registries
We currently reference a couple images from quay.io and docker.io.  These should be brought into our ACR instance and referenced that way instead.  Update the instructions to include this step and changes.

ckittel:
* Add Traefik to Ingress Controller
* Didn't add the demo app, as that is from MCR
* You may notice the `docker.io` vs `registry.hub.docker.com` -- they both resolved the same in practice and went with consistency in the docs

This closes #98


10-validation.md
```diff
@@ -133,7 +133,7 @@ If you configured your third-party images to be pulled from your Azure Container
    | where OperationName == 'Pull'
    ```
 
-1. You should see logs for kured, CSI, flux, and memcached.  You'll see multiple for some as the image was pulled to multiple nodes to satisfy ReplicaSet/DaemonSet placement.
+1. You should see logs for CSI, flux, kured, memcached, and traefik.  You'll see multiple for some as the image was pulled to multiple nodes to satisfy ReplicaSet/DaemonSet placement.
```
hallihan, line 136: +1 for listing of logs sorted in alphabetical order 👍

hallihan (approved):
Looks good, :shipit:

10-validation.md
ckittel: Don't want to imply preference for any one technology :)

Issue #104, opened 2020-11-02
Upgrade Azure AD Pod Identity bits


ckittel:
Do an upgrade pass on AAD pod identity - to 1.6.3 (or whatever is latest)

* Ensure we're pulling from the new `mcr.microsoft.com/oss/azure/aad-pod-identity` location as well.

ckittel:
Also consider moving to the new AAD Pod Identity add-on

ckittel:
Not going to the AAD Pod Identity add-on with this issue. Opened a new issue to track that.

#111

PR #110, opened 2020-11-13
Update to latest AAD Pod Identity

ckittel:
* Update from 1.6.1 to 1.6.3 (no breaking changes wrt this baseline implementation)
* move ClusterRoleBinding to v1 api
* Moved exceptions into the AAD Pod Identity yaml to prevent race conditions from flux processing expectations before AAD Pod Identity CRDs were installed.
* Tightened up pod: removed unneeded mount and unneeded privileged security context.

This addresses #104


cluster-baseline-settings/aad-pod-identity/aad-pod-identity.yaml
```diff
@@ -201,32 +204,25 @@ spec:
...
+      dnsPolicy: ClusterFirstWithHostNet
       volumes:
       - hostPath:
           path: /run/xtables.lock
           type: FileOrCreate
         name: iptableslock
       containers:
       - name: nmi
-        image: "mcr.microsoft.com/k8s/aad-pod-identity/nmi:1.6.1"
+        image: "mcr.microsoft.com/oss/azure/aad-pod-identity/nmi:v1.6.3"
```
ferantivero, line 215: should this be also `imported`?

ferantivero: if yes, but since this is temporary `yaml` we could leave as-is.

ferantivero (approved):
LGTM 🚢

cluster-baseline-settings/aad-pod-identity/aad-pod-identity.yaml
ckittel: Nope, doesn't need to be "mcr" is a "safe" public repo to pull from.  Doesn't throttle.  And, since we'll be moving to AAD as an add-on soon enough, it'll go away as you said.

PR #112, opened 2020-11-21
Update allowed regions with new AZ supported list

ckittel:
... and their paired region.  All other used SKUs available as well in those regions.


ckittel:
Going to merge in, but @v-fearam if you have any comments, please do let me know.

PR #114, opened 2020-11-21
AAD Pod Identity 1.7.0

### LINKED ISSUE #113: Update Azure AD AddOn to 1.7.0
Latest Azure AD Addon has been released, upgrade to 1.7.0 (or whatever is newer once this is picked up)

https://github.com/Azure/aad-pod-identity/releases/tag/v1.7.0

ckittel:
This update
* brings the bits to 1.7.0
* declairs the runAsUser to appease CIS benchmark scanning tools -- matches already existing requirement (both run as root already as one interfaces with `iptables` and the other interfaces with `/etc/kubernetes/azure.json`)
* new kubelet-config mount necessary since kubenet and AAD Pod Identity interactions are a bit different now.

I've tested an end to end run of this.

Fixes #113

### DIFF
--- a/cluster-baseline-settings/aad-pod-identity/aad-pod-identity.yaml
+++ b/cluster-baseline-settings/aad-pod-identity/aad-pod-identity.yaml
@@ -210,9 +210,12 @@ spec:
           path: /run/xtables.lock
           type: FileOrCreate
         name: iptableslock
+	    - hostPath:
+          path: /etc/default/kubelet
+        name: kubelet-config
       containers:
       - name: nmi
-        image: "mcr.microsoft.com/oss/azure/aad-pod-identity/nmi:v1.6.3"
+        image: "mcr.microsoft.com/oss/azure/aad-pod-identity/nmi:v1.7.0"
         imagePullPolicy: Always
         args:
           - "--node=$(NODE_NAME)"
@@ -223,12 +226,16 @@ spec:
               fieldRef:
                 fieldPath: spec.nodeName
         securityContext:
+	        runAsUser: 0
           capabilities:
             add:
             - NET_ADMIN
         volumeMounts:
         - mountPath: /run/xtables.lock
           name: iptableslock
+	      - mountPath: /etc/default/kubelet
+          name: kubelet-config
+          readOnly: true
         livenessProbe:
           httpGet:
             path: /healthz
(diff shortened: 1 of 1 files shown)


cluster-baseline-settings/aad-pod-identity/aad-pod-identity.yaml
```diff
@@ -223,12 +226,16 @@ spec:
               fieldRef:
                 fieldPath: spec.nodeName
         securityContext:
+          runAsUser: 0
```
ferantivero, line 229: this value was requested to be higher than 10k before to avoid conflicts with host user tables IIRC.

ckittel: Sorry, not sure what you mean. Before what? It wasn't defined before. This runs as root by design.

ckittel: To expand on the above...

In the prior version of this, `runAsUser` was not needed to be set, because the container itself set the user to root.  Now for added "compliance" -- the container is set to run as a non-root user, in its defination, however, that means this "decision" to run as root is pushed up to `securityContext` -- in practice, it needs to be run as root -- and the new method of this going forward is to designate it as such here, vs in the container.  This was one of the 1.7.0 changes.

I don't know what host user table conflict you're talking about here, unfortunatly.  Since this worked end-to-end, I'm not sure I know how to test for the conflict you mention.

ckittel: @ferantivero -- any chance you'll get a look at this yet today?  I'm thinking of starting on the AKS 1.19 work over this extended weekend, and would love to base it on this if this is good to go.

ferantivero: hey there, sorry for the delay in here, for some reason I haven't seen the replies.

> "Before what?"

In the past, we had to run some advisory tooling in our clusters. As a result from it, we were requested to change `runAsUser` values to something higher than `10k`. Otherwise you might run into conflicts with the host's user tables and it is not desired for security reason.

Something I didn't notice in here is that our desired is to run this "as root". If that is the case, please let's discard this comment.

ferantivero (approved):
LGTM 🚀

PR #121, opened 2020-12-02
update minimum and scaled subnet size

ferantivero:


ckittel (approved):
Doh, good catch!  I guess something was a fn() of that first value afterall 👍   Thanks!

Issue #122, opened 2020-12-03
missing metric (oomKilledContainerCount ) when creating cluster with ARM template


mracfa:
Hi,
On step 5 (05-aks-cluster) it fails the creation of the cluster if we use the ARM template due to this missing metric: **oomKilledContainerCount**:

```bash
maabre@MAABRE-BOOK2:/mnt/c/Users/maabre/wls_aks/git/aks-secure-baseline$ az deployment group create --resource-group rg-bu0001a0008 --template-file cluster-stamp.json --parameters targetVnetResourceId=$TARGET_
VNET_RESOURCE_ID k8sRbacAadProfileAdminGroupObjectID=$K8S_RBAC_AAD_PROFILE_ADMIN_GROUP_OBJECTID k8sRbacAadProfileTenantId=$K8S_RBAC_AAD_PROFILE_TENANTID appGatewayListenerCertificate=$APP_GATEWAY_LISTENER_CERT
IFICATE aksIngressControllerCertificate=$AKS_INGRESS_CONTROLLER_CERTIFICATE_BASE64
Deployment failed. Correlation ID: 009f4644-47b3-4846-a88d-9db844f68583. {
  "Code": "BadRequest",
  "Message": "The following metric name(s) were not found: oomKilledContainerCount. Please note that for custom metrics, the relevant metric namespace must be specified. Activity ID: 93208c19-930f-425a-a71b-e182a78937a9."
}
```

Can you please check?

ckittel:
Interesting.  We hit that with two other ones as well and had to remove them from the deployment.  I wonder if it's regional.  What region were you headed into?  To bypass this, simply remove the `Microsoft.Insights/metricAlerts` resource for that one from `cluster-stamp.json`.  I know alerts are technically `global` for region, but there might be something else going on preventing the deployment.  I bet if you drop this from the template, and after this is deployed, you go into the portal "<cluster>->Insights->Recommended alerts" and find the (now) three that would be sitting at disabled, you could enable them from there.  There seems to be some disconnect between how the portal gets away with creating these vs ARM template.  Let me know what you find, I'd love to get to the bottom of this.

ckittel:
On the region topic, I winder if you'll find you're in a region that is not on the following list?
<https://docs.microsoft.com/azure/azure-monitor/platform/metrics-custom-overview#supported-regions>  -- Interestingly, even `eastus2` had an issue with two of the alerts when I deployed via ARM template -- so not really sure yet what's going on.

[...]

Issue #117, opened 2020-12-01
Add Default Metric Alerts


ckittel:
Wanted to add these with the 1.19 upgrade, but currently they are erroring out when created via ARM template.  When done via the portal (via the preview experience), they work.  Cannot create these either via the portal in the direct mertic alert experience.  Circle back and add them once they can be created via template.

```json
     {
            "type": "Microsoft.Insights/metricAlerts",
            "location": "global",
            "apiVersion": "2018-03-01",
            "name": "[concat('Node CPU utilization high for ', variables('clusterName'), ' CI-1')]",
            "dependsOn": [
                "[resourceId('Microsoft.ContainerService/managedClusters', variables('clusterName'))]",
                "[resourceId('Microsoft.OperationsManagement/solutions',variables('containerInsightsSolutionName'))]"
            ],
            "properties": {
                "actions": [],
                "criteria": {
                    "allOf": [
                        {
                            "criterionType": "StaticThresholdCriterion",
                            "dimensions": [
                                {
                                    "name": "host",
                                    "operator": "Include",
                                    "values": [
                                        "*"
                                    ]
                                }
                            ],
                            "metricName": "cpuUsagePercentage",
                            "metricNamespace": "Insights.Container/nodes",
                            "name": "Metric1",
                            "operator": "GreaterThan",
                            "threshold": 80.0,
                            "timeAggregation": "Average"
                        }
                    ],
                    "odata.type": "Microsoft.Azure.Monitor.SingleResourceMultipleMetricCriteria"
                },
                "description": "Node CPU utilization across the cluster.",
                "enabled": true,
                "evaluationFrequency": "PT1M",
                "scopes": [
                    "[resourceId('Microsoft.ContainerService/managedClusters', variables('clusterName'))]"
                ],
                "severity": 3,
                "targetResourceType": "microsoft.containerservice/managedclusters",
                "windowSize": "PT5M"
            }
        },
        {
            "type": "Microsoft.Insights/metricAlerts",
            "location": "global",
            "apiVersion": "2018-03-01",
            "name": "[concat('Node working set memory utilization high for ', variables('clusterName'), ' CI-2')]",
            "dependsOn": [
                "[resourceId('Microsoft.ContainerService/managedClusters', variables('clusterName'))]",
                "[resourceId('Microsoft.OperationsManagement/solutions',variables('containerInsightsSolutionName'))]"
            ],
            "properties": {
                "actions": [],
                "criteria": {
                    "allOf": [
[...]

PR #123, opened 2020-12-03
Temporarily disable auto metric alert creation

ckittel:
Due to feedback from customers and internal MSFT employees, there appears to be a race condition between cluster creation and these metric alerts being able to be created, which causes failures with random ones.  For now disable them.  I'm going to meet with the PG on these and see if there is a better solution than having this happen "minutes" after cluster creation.  Also added the other two back in that I had removed for the same reason (node CPU and node Memory -- both at 80% over 5 minutes -- default values).

I think the reason I didn't run into this in all my end-to-end testing was when I was stopped by the race condition, the time took to fix the problem (removing the items) caused enough time for the rest of the metric namespace additions to propagate.

Context: #122 and #117


ferantivero (approved):
LGTM 🚀

PR #124, opened 2020-12-03
Reenable the OOTB Metric Alerts

ckittel:
Pass along a "skipMetricValidation" request with the resources, this allows the metric alert to be created without forcing the user to wait for the metric to exist.

Also disabled the PV alert as... since we don't have PV, it's not cost effective to enable it and shows selective usage of alerts based on the shape/needs of the cluster.


ferantivero (approved):
w00t | LGTM 🚀

PR #127, opened 2020-12-07
Update to 0.0.10 AKV-CSI

ckittel:
* Update to latest version of AKV-CSI Integration
* Update storage driver API to remove deprecation warning
* Grab container images from MCR for AKV so we don't need to import them.
* Lifecycle for registrar is removed -- new image is distroless (which means it doesn't have a shell) and now this is handled in the container via other means
* `privileged` now `false` with the new `mountPropagation`.


hallihan (approved):
Lgtm, :shipit:

PR #132, opened 2020-12-10
Adding a PDF render of the HTML presentation

dcasati:
This is a PDF version of the HTML (remark.js) presentation.


ckittel (approved):

ckittel:
I think I'm gonig to hold off at this point, I'm making some wording/flow updates, and this will be out of sync with those updates. Thank you though!

PR #136, opened 2021-01-13
Add a second spoke in the same region

ferantivero:
add a second spoke which is meant to be AKS cluster development oriented


networking/spoke-BU0001A0042.json
```diff
@@ -0,0 +1,256 @@
+{
```
ferantivero, line 1: change the naming taking into account that it is going to be the same application in multiple regions `spoke-BU0001A0042-00` and `spoke-BU0001A0042-01`.

Initially it is ok use same application, same region but the identifier needs to be differ.

networking/spoke-BU0001A0042.json
```diff
@@ -0,0 +1,256 @@
...
+            "apiVersion": "2020-07-01",
+            "name": "[variables('clusterVNetName')]",
+            "location": "[parameters('location')]",
+            "dependsOn": [
+                "[resourceId('Microsoft.Network/routeTables', variables('routeTableName'))]"
+            ],
+            "properties": {
+                "addressSpace": {
+                    "addressPrefixes": [
+                        "10.241.0.0/16"
```
ferantivero, line 76: change this to do not conflict with neither `baseline` and `regulated`:

baseline: 10.240.X.X
regulated: 10.241.X.X + 10.242.X.X 
multicluster: 10.243.X.X + 10.244.X.X

networking/spoke-BU0001A0042.json
```diff
@@ -0,0 +1,256 @@
...
+                "virtualNetwork": {
+                    "id": "[resourceId('Microsoft.Network/virtualNetworks', variables('clusterVNetName'))]"
+                },
+                "registrationEnabled": false
+            }
+        },
+        {
+            "type": "Microsoft.Network/publicIpAddresses",
+            "apiVersion": "2020-07-01",
+            "name": "[concat('pip-', variables('orgAppId'), '-00')]",
```
ferantivero, line 230: change this to
`pip-bu0001a0042-agw-00`
`pip-bu0001a0042-agw-01`

networking/spoke-BU0001A0042.json
```diff
@@ -0,0 +1,256 @@
...
+                "southeastasia"
+            ],
+            "metadata": {
+                "description": "The spokes's regional affinity. All resources tied to this spoke will also be homed in this region.  The network team maintains this approved regional list which is a subset of zones with Availability Zone support."
+            }
+        }
+    },
+    "variables": {
+        "orgAppId": "BU0001A0042",
+        "clusterVNetName": "[concat('vnet-hub-spoke-', variables('orgAppId'), '-00')]",
```
ferantivero, line 37: templetize this:

ask our users to pass the minimum amount of required args. If possible without confusing them not even the full address prefix.

parameter: appInstanceNumber [3,4] --> add TODO consider mutating this into region instead when we go multi-region.

Address Prefix: 10.24<3>
pip: -0<3> 

Address Prefix: 10.24<4>
pip: -0<4>

networking/spoke-BU0001A0042.json
ferantivero: done, but addressed with pip-agw-<instance> instead of

networking/spoke-BU0001A0042.json
ferantivero: done

PR #137, opened 2021-01-15
Add cluster bu0001a0042

ferantivero:
add a second cluster stamp using same app gw and aks ingress controller certs

### DIFF
--- a/05-aks-cluster.md
+++ b/05-aks-cluster.md
@@ -1,138 +1,42 @@
 # Deploy the AKS Cluster
 
-Now that the [hub-spoke network is provisioned](./04-networking.md), the next step in the [AKS secure Baseline reference implementation](./) is deploying the AKS cluster and its adjacent Azure resources.
+Now that the [hub-spoke network is provisioned](./04-networking.md), the next step in the [AKS secure Baseline reference implementation](./) is deploying the AKS clusters and its adjacent Azure resources.
 
 ## Steps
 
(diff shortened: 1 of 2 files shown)


cluster-stamp-bu0001a0008.json
```diff
@@ -474,19 +474,41 @@
                     {
                         "Name": "apw-routing-rules",
                         "properties": {
-                            "RuleType": "Basic",
+                            "RuleType": "PathBasedRouting",
```
ferantivero, line 477: go back to basic. It is not the same app + from validation standpoint this can be done using multiple dns A records to different IP address. Therefore, it is going to randomly access the different gateways.

cluster-stamp-bu0001a0008.json
```diff
@@ -474,19 +474,41 @@
                     {
```
ferantivero, line 474: templetized same as the spoke. Most probably using the the application instance number.

cluster-stamp-bu0001a0008.json
ferantivero: we could use the remote address from devtools as a option (Edge).

ferantivero: done - go back to basic. 

Pending create validation steps

cluster-stamp-bu0001a0008.json
ferantivero: done

PR #148, opened 2021-02-05
Multi cluster - Azure Front Door Narrative

v-fearam:
- Adding the instruction and narrative in order to include Azure Front Door
- Doing some fixing on script
- Improving the certification generation explanation and making easy to change the name in case of global DNS collision.


ferantivero (changes requested):
incorporate these changes to  the main README file.

ferantivero (approved):

PR #151, opened 2021-02-08
adapt validation step

ferantivero:


v-fearam:
Please take a look https://github.com/mspnp/aks-secure-baseline/pull/148, we are doing the same

PR #153, opened 2021-02-09
Limiting Application Gateway Access

v-fearam:
- Limiting Application Gateway Access (only allow Front Door request)
- Small Fix -

### DIFF
--- a/networking/spoke-BU0001A0042.json
+++ b/networking/spoke-BU0001A0042.json
@@ -182,12 +182,12 @@
             "properties": {
                 "securityRules": [
                     {
-                        "name": "Allow443InBound",
+                        "name": "AllowOnlyFrontDoor",
                         "properties": {
-                            "description": "Allow ALL web traffic into 443. (If you wanted to allow-list specific IPs, this is where you'd list them.)",
+                            "description": "Allow only Front Door Access. (If you wanted to allow-list specific IPs, this is where you'd list them.)",
                             "protocol": "Tcp",
                             "sourcePortRange": "*",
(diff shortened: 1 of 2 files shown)


networking/spoke-BU0001A0042.json
```diff
@@ -182,12 +182,12 @@
...
                     {
-                        "name": "Allow443InBound",
+                        "name": "AllowOnlyFrontDoor",
                         "properties": {
-                            "description": "Allow ALL web traffic into 443. (If you wanted to allow-list specific IPs, this is where you'd list them.)",
+                            "description": "Allow only Front Door Access. (If you wanted to allow-list specific IPs, this is where you'd list them.)",
                             "protocol": "Tcp",
                             "sourcePortRange": "*",
-                            "sourceAddressPrefix": "Internet",
+                            "sourceAddressPrefix": "AzureFrontDoor.Backend",
```
ferantivero, line 190: I think it is actually reducing the surface of attack based on this new change in our user's NSG. Although please let's take into account that any other AFD instances out there could contact the App Gws since there is no way to make a rule for an instance in particular.

I think what is most important is that this is adding some busy work but not letting bad actors to open a browser and bypass our infrastructure deliberately.

+1 to keep this rule, but I would like to double check this with @ckittel

ckittel: Yeah, we can (and should) keep it.  Because it is accurate and is a form of self-documentation.  Just doesn't provide a lot of protection -- so we shouldn't necessarily call it out as a security measure other than one of many defense in depth. solutions.

PR #157, opened 2021-02-10
[Muti Cluster] Changing cert names

v-fearam:


02-ca-certificates.md
```diff
@@ -21,23 +21,23 @@ Following the steps below you will result the certificate needed for Azure Appli
 
    Create a CA certificate for each Azure Application Gateway. You can use your company domain or try get a certificate for each domain using [Azure Subdomain Certificates Generation](./certificate-generation/README.md).
 
-   :warning: We called bicycle3 and bicycle4 each subdomain, but the DNS values could be not available. In that case, you can change the following values.
+   :warning: We are waiting for two certificates in the following example (Please, pay attention the regions). We called bu0001a004203 (eastus2) and bu0001a004204 (westus2) each subdomain, but the DNS values could be not available. In that case, you can change the following values.
 
    ```bash
-   export CLUSTER_SUBDOMAIN_03=bicycle3
-   export CLUSTER_SUBDOMAIN_04=bicycle4
+   export CLUSTER_SUBDOMAIN_03=bu0001a004203
```
ferantivero, line 27: please let's use another repeatable but randomly generated unique string. This is because it is revealing internal business implementation details from public endpoints.

Can we generate them directly in the ARM?

ferantivero (changes requested):

02-ca-certificates.md
v-fearam: I have changing for a combination of azure username and region1 and region2. 
I needed to delete '.' from user name, and the azure storage account doesn't support . and -

02-ca-certificates.md
```diff
@@ -21,14 +21,15 @@ Following the steps below you will result the certificate needed for Azure Appli
...
    Create a CA certificate for each Azure Application Gateway. You can use your company domain or try get a certificate for each domain using [Azure Subdomain Certificates Generation](./certificate-generation/README.md).
 
-   :warning: We are waiting for two certificates in the following example (Please, pay attention the regions). We called bu0001a004203 (eastus2) and bu0001a004204 (westus2) each subdomain, but the DNS values could be not available. In that case, you can change the following values.
+   :warning: We are waiting for two certificates in the following example (Please, pay attention the regions where the certs are needed). We called CLUSTER_SUBDOMAIN_03 and CLUSTER_SUBDOMAIN_04 each subdomain. You can change the name if you desire.
 
    ```bash
-   export CLUSTER_SUBDOMAIN_03=bu0001a004203
-   export CLUSTER_SUBDOMAIN_04=bu0001a004204
+   USER_NAME=$(az ad signed-in-user show --query 'mailNickname' -o tsv)
+   export CLUSTER_SUBDOMAIN_03=${USER_NAME//.}-region1
```
ferantivero, line 28: I would let @ckittel decide what is better here. I think user names are ok. But any desire to use a a more random and repeatable (as far as I can see) solution using a number instead?

`export SUBDOMAIN_REGION1=$(bash -c 'RANDOM=$(az ad user show --id <e-Mail> --query immutableId -o tsv); echo region1-$RANDOM$RANDOM$RANDOM')`

it generate something like:

```output
region1-2221110329407
```

02-ca-certificates.md
ckittel: Can you say more about what you mean by "azure username" -- we don't want any business or personal data leaking out as an implementation detail.  Region and service is usually bad enough (but often unavoidable), is something like `uniqueString(ip resource id)` not sufficient?

02-ca-certificates.md
ferantivero: based on @ckittel comments above, I think it might safe to proceed with this solution.

ckittel: yeah, just as long as it's natural and tied to the lifecycle of a _resource_ ... not a human.  Human relationships end with organizations (sometimes poorly) -- having some sort of value based on `employeeWeWishToForget` would be hilarious, but also terrible if someone actually does this literally ... :)  -- let's keep the humans out, and the robots/resources in :)

02-ca-certificates.md
v-fearam: 1- We like to avoid dns collision, so we need something unrelared to resource group name. The user name is not private information, and it is unique for demo propose(in real deploy your company use your domain and names. In the narrative says that you could change, we can say must change o put nothing and you are force to choose one),
2- We need the cert before deploying the cluster, we can use a ip derivated value after spoke deploy and move the cert generation between spoke and cluster. It takes time, but sounds like a good idea.

I like to avoid select the name and force the user to select one, @ferantivero @ckittel ?

ferantivero: so based on our discussion the final decision in here is to generate this via ARM, and we plan to change the timing for the cert generation later on between networking and the cluster creation.

v-fearam: ready a new version to take a look

ferantivero (approved):
LGTM

PR #158, opened 2021-02-12
[Multi Cluster] Extract ACR as regional but shared service

ferantivero:
closes: #349791

- modify sections to introduce a new one for cluster prerequisites and shared service instances like ACR.
- deploy shared services using a new stamp
- extract shared resources from cluster-stamp
   - ACR and its private dns zone
   - log analytics to centralize the logs
- deploy to paired regions
   - `cluster03` -> `eastus2`
   - `cluster04` -> `centralus`


v-fearam (approved):

shared-svcs-stamp.json
```diff
@@ -0,0 +1,287 @@
+{
+    "$schema": "https://schema.management.azure.com/schemas/2019-04-01/deploymentTemplate.json#",
+    "contentVersion": "0.0.0.5",
+    "parameters": {
+        "location": {
```
ferantivero, line 5: add brazil as an option.

cluster-stamp.json
```diff
@@ -725,79 +769,47 @@
...
+                            }
+                        }
+                    ]
                 }
-            ]
+            }
         },
         {
-            "name": "PodFailedScheduledQuery",
+            "name": "[concat('PodFailedScheduledQuery','-',variables('clusterName'))]",
```
ferantivero, line 802: see if we can use the new way of defining queries and adding scopes later + using the dimension feature. Take a look from regulated.

cluster-stamp.json
```diff
@@ -848,160 +860,12 @@
                 "description": "All azure advisor alerts"
```
ferantivero, line 860: investigate whether we can move this to shared resources stamp and enroll resources later (ideal) or remove the scope and keep this at the subs lvl (second best). Otherwise make them unique.

cluster-stamp.json
```diff
@@ -274,12 +315,12 @@
                     "name": "Microsoft.Insights/default",
                     "type": "providers/diagnosticSettings",
                     "apiVersion": "2017-05-01-preview",
+                    "location": "[parameters('logAnalyticsWorkspaceLocation')]",
```
ferantivero, line 318: re-validate for this resource type whether location is required or not.

shared-svcs-stamp.json
```diff
@@ -0,0 +1,287 @@
...
+            }
+        }
+    },
+    "variables": {
+        "subRgUniqueString": "[uniqueString('aks',subscription().subscriptionId, resourceGroup().id)]",
+
+        "logAnalyticsWorkspaceName": "[concat('la-', variables('subRgUniqueString'))]",
+        "defaultAcrName": "[concat('acraks', variables('subRgUniqueString'))]",
+
+        "acrPrivateDnsZonesName": "privatelink.azurecr.io"
```
ferantivero, line 66: how is going to resolve the private DNS the pull request of a container image. Should we be specific from what region this is downloaded?

shared-svcs-stamp.json
ferantivero: Done | Addressed from [9ce2c](https://github.com/mspnp/aks-secure-baseline/pull/158/commits/9ce2c466ce890f57b6985214e9f7a8dbe8fe6fa3)

cc/ @ckittel  I think we included [all products](https://azure.microsoft.com/en-us/global-infrastructure/services/?regions=brazil-south,non-regional,brazil-southeast,us-east-2,us-central&products=storage,azure-bastion,frontdoor,azure-firewall,virtual-network,load-balancer,application-gateway,key-vault,active-directory,monitor,container-registry,kubernetes-service)

cluster-stamp.json
ferantivero: Done | Addressed from [06c29](https://github.com/mspnp/aks-secure-baseline/pull/158/commits/06c293627751aeadccc87849fa55cbb2e4434444)

diagnostic does not require the location as discussed while alerts will fail with being on the same location as the centralized LA

```json
"error": {
    "code": "Workspace not found. Please note that this may be due to alert rule location, which must be the same as the workspace location.",
    "message": {
      "lang": "en-US",
      "value": "Workspace not found. Please note that this may be due to alert rule location, which must be the same as the workspace location."
    }
  }
```

cluster-stamp.json
ferantivero: Done | Addressed from 01518b87fcb4f2a51cab216612de5fab63362167

cc/ @ckittel I had to go with the third option. I know it wast the least desired 😞 but based on [docs](https://docs.microsoft.com/en-us/azure/templates/microsoft.insights/activitylogalerts#ActivityLogAlert) I would say is our only option. 

image

Btw it seems you could dup the name without issues, please take a look at the screenshots below for two different rg:

image

image

ferantivero: hold on with this. Just found an example in another object giving the scope the subscription id while requires "resource ids". Just testing it to see if it works.

ferantivero: it did work, and it now addressed from [96373](https://github.com/mspnp/aks-secure-baseline/pull/158/commits/96373535c9f271774b7434489d692ac30ae998ae)

cluster-stamp.json
ferantivero: yes, we can and it is now done |  Addressed from [c7bf4](https://github.com/mspnp/aks-secure-baseline/pull/158/commits/c7bf4e10344dd86c5d55eb959538d5388c321d59)

cluster-stamp.json
ferantivero: Although above error is true, the new way of making `scheduledQueryRules` allows you to create this without specifying the LA resource id and with that we can avoid passing the location. 

Rosolved via https://github.com/mspnp/aks-secure-baseline/pull/158#discussion_r583693391

shared-svcs-stamp.json
ferantivero: I plan to tackle this as part of a different PR/Task.

PR #159, opened 2021-02-12
added steps to know current roles & permissions

Mahesh-MSFT:
updated prerequisites to let users know what their current role is. also, modified link to make differentiation between roles and permissions.

### DIFF
--- a/01-prerequisites.md
+++ b/01-prerequisites.md
@@ -10,12 +10,23 @@ This is the starting point for the instructions on deploying the [AKS Secure Bas
    >
    > * [Contributor role](https://docs.microsoft.com/azure/role-based-access-control/built-in-roles#contributor) is _required_ at the subscription level to have the ability to create resource groups and perform deployments.
    > * [User Access Administrator role](https://docs.microsoft.com/azure/role-based-access-control/built-in-roles#user-access-administrator) is _required_ at the subscription level since you'll be performing role assignments to managed identities across various resource groups.
+   >
+   > Run following CLI commands to understand current roles associated with Azure account.
+   >
+   >```azurecli-interactive
+   > az login
+   >```
+   >
+   >```azurecli-interactive
+   > az role assignment list --all --assignee <YOUR-LOGIN-ACCOUNT> --output json --query '[].{roleDefinitionName:roleDefinitionName, scope:scope}'
+   >`
+
 
 1. An Azure AD tenant to associate your Kubernetes RBAC Cluster API authentication to.
 
    > :warning: The user or service principal initiating the deployment process _must_ have the following minimal set of Azure AD permissions assigned:
    >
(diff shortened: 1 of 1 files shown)


01-prerequisites.md
```diff
@@ -10,12 +10,23 @@ This is the starting point for the instructions on deploying the [AKS Secure Bas
...
    > * [User Access Administrator role](https://docs.microsoft.com/azure/role-based-access-control/built-in-roles#user-access-administrator) is _required_ at the subscription level since you'll be performing role assignments to managed identities across various resource groups.
+   >
+   > Run following CLI commands to understand current roles associated with Azure account.
+   >
+   >```azurecli-interactive
+   > az login
+   >```
+   >
+   >```azurecli-interactive
+   > az role assignment list --all --assignee <YOUR-LOGIN-ACCOUNT> --output json --query '[].{roleDefinitionName:roleDefinitionName, scope:scope}'
```
brk3, line 21: I think this may have potential to cause confusion for the following reasons:

1. Contributor role _or higher_ is required. E.g. I have a classic administrator role on my subscription, which works fine for the purposes of this guide, but results in the above command returning empty.
2. We need both RBAC roles and Azure AD roles (User Access Administrator) in place. This command only checks the RBAC piece.

01-prerequisites.md
```diff
@@ -10,12 +10,23 @@ This is the starting point for the instructions on deploying the [AKS Secure Bas
...
+   >`
+
 
 1. An Azure AD tenant to associate your Kubernetes RBAC Cluster API authentication to.
 
    > :warning: The user or service principal initiating the deployment process _must_ have the following minimal set of Azure AD permissions assigned:
    >
-   > * Azure AD [User Administrator](https://docs.microsoft.com/azure/active-directory/users-groups-roles/directory-assign-admin-roles#user-administrator-permissions) is _required_ to create a "break glass" AKS admin Active Directory Security Group and User. Alternatively, you could get your Azure AD admin to create this for you when instructed to do so.
+   > * Azure AD [User Administrator Permissions](https://docs.microsoft.com/azure/active-directory/users-groups-roles/directory-assign-admin-roles#user-administrator-permissions) are _required_ to create a "break glass" AKS admin Active Directory Security Group and User. Alternatively, you could get your Azure AD admin to create this for you when instructed to do so.
```
brk3, line 29: I think what's already written is correct and doesn't need to be changed. It reads clearer to me that I need to apply a specific role (User Administrator), rather than the collection of permissions it encompasses.

ckittel: Agreed.

ckittel:
@Mahesh-MSFT -- I'm going to go ahead and close this PR for now. If we want to take another stab at adding additional clarity here, let's build in the thoughts from @brk3 and either reopen this PR or maybe a new one.  Thanks!

PR #167, opened 2021-03-01
[Multi Cluster] add narrative around LA for it to be managed at scale

ferantivero:
closed: #349799

add narrative around Azure Log Analytics and what was the process of choosing a `Centralized` solution over a `Decentralized`


05-cluster-prerequisites.md
```diff
@@ -27,7 +27,16 @@ Following the steps below will result in the provisioning of the shared Azure re
...
+   >
+   > - Azure Container Registry
+   > - Azure Application Gateway
+   > - Azure Key Vault
+   >
+   > In the future, the app team is considering to enforce the Azure resources to send their logs with an Azure Policy as well as  granting different users access rights to keep the data in isolation using Azure RBAC, something that is possible within a single workspace.
+
+   > :bulb:  Azure Log Analytics can be modeled in different ways depending on your orgnization needs. It can be `Centralized` as in this reference implementation, or `Decentralized` or a combination of both, which is known as `Hybrid`. Azure Log Analytics workspaces are in a geographic location for data storage, so consider for High Availability a distributed solution as the recommended approach instead. You can opt for a `Centralized` solution like in here but you need to be sure that the data residency is not going to be an issue, and be aware that data transfer may pay fees while traveling from one region to the another.
+
```
ckittel, line 39: Here are a couple nits -- nothing critical -- take/leave what you wish.
```suggestion
   > :book: The app team is provisioning an Azure Container Registry instance that is shared by their AKS clusters as well as their [Azure Monitor logs solution](https://docs.microsoft.com/azure/azure-monitor/logs/design-logs-deployment). The app team is creating multiple clusters for its new workload (Application Id: a0042). This array of clusters is a multi-region infrastructure solution composed by multiple Azure resources that regularly emit logs to Azure Monitor. All the collected data is stored in a centralized Log Analytics workspace for the ease of operations after confirming there is no need to split workspaces due to scale. The app team estimates that the ingestion rate is going to be less than `6GB/minute`, so they expect not to be throttled as this is supported by the default rate limit. If it was required, they could grow by changing this setup eventually. In other words, the design decision is to create a single Azure Log Analytics workspace instance in the `eastus2` region, and that is going to be shared among their multiple clusters. Additionally, there is no business requirement for a consolidated cross business units view at this moment, so the centralization is great option for them. Something the app team also considered while making a final decision is the fact that migrating from a _centralized_ solution to a  _decentralized_ one can be much easier than doing it the other way around. As a result, the single workspace being created is a workload specific workspace, and these are some of the Azure services sending data to it:
   >
   > - Azure Container Registry
   > - Azure Application Gateway
   > - Azure Key Vault
   >
   > In the future, the app team is considering to enforce the Azure resources to send their Diagnostics logs with an Azure Policy as well as granting different users access rights to keep the data in isolation using Azure RBAC, something that is possible within a single workspace.

   > :bulb:  Azure Log Analytics can be modeled in different ways depending on your organizational needs. It can be _centralized_ as in this reference implementation, or _decentralized_ or a combination of both, which is known as _hybrid_. Azure Log Analytics workspaces are in a geographic location for data storage; consider, for high availability, a distributed solution as the recommended approach instead. If you opt for a _centralized_ solution, you need to be sure that the geo data residency is not going to be an issue, and be aware that cross-region data transfer costs will apply.
```

ckittel, line 39: That's good information.  I like that you called out what the options are and considerations for each.  Any desire to have a Azure Container Registry break-down conversation like you did here for LA?  If so, I think you have a good model to work from with this.

ckittel (approved):
<3

05-cluster-prerequisites.md
ferantivero: <3 thanks for the feedback @ckittel, yes, I plan going over ACR as part my next task which is General Considerations around Management at Scale.

PR #170, opened 2021-03-02
[Multi Cluster] Add narrative for Azure Container Registry

ferantivero:
related: #374114

add narrative for Azure Container Registry management at scale

 - Costs considerations 
 - Fail over
 - Networking
 - A starting point for ACR and _Availability Zones_
 - More


05-cluster-prerequisites.md
```diff
@@ -35,14 +35,25 @@ Following the steps below will result in the provisioning of the shared Azure re
...
...ens to live in different regions. Although managing registry instances per region individually will mitigate the risk of having a region down, this approach won’t be falling back automatically nor replicating images by requiring in both cases manual intervention and/or additional procedures.  That is the reason why, the team is now up for the `Premium` tier that offers [Geo-Replication](https://docs.microsoft.com/en-us/azure/container-registry/container-registry-geo-replication) as a built-in feature giving them the ability to share a single registry, more availability, and at same time reduce the network latency.  The app team plan to geo replicate the registries to the same regions where their AKS clusters are going to be deployed (`East US 2` and `Central US`), as this is the recommendation to optimize the DNS resolution. They pay close attention to the replicated regions, as they want no more regions than needed since the business unit (bu0001) incurs `Premium` SKU registry fees for each region they desire to geo-replicate. Under this configuration, they will pay *two* times `Premium` per month to get region proximity to ensure no extra network egress fees from distant regions.
```
ckittel, line 48: > This requirement comes in hand with distributing the exact same container images as this is for tween clusters that happens to live..

> This is turn for the Azure Container Registry with the additional problematic that container images can be heavy artifacts to deal with, so they a

Something seems a bit off in these two sentences to me. I think it's just wording, not message.




```suggestion
  > :book: The app team is starting to lay down the groundwork for a high availability, and they know that centralizing Azure resources might introduce single point of failures. Therefore, the app team is tasked to assess how the resources could be shared efficiently without losing reliability. This is turn for the Azure Container Registry with the additional problematic that container images can be heavy artifacts to deal with, so they are considering having presence in multiple regions as well. This requirement comes in hand with distributing the exact same container images as this is for tween clusters that happens to live in different regions. Although managing registry instances per region individually will mitigate the risk of having a region down, this approach won’t be falling back automatically nor replicating images by requiring in both cases manual intervention and/or additional procedures.  That is the reason why, the team selected the **Premium** tier that offers [Geo-Replication](https://docs.microsoft.com/en-us/azure/container-registry/container-registry-geo-replication) as a built-in feature; giving them the ability to share a single registry, higher availability, and at same time reducing network latency. The app team plans to geo-replicate the registries to the same regions where their AKS clusters are going to be deployed (`East US 2` and `Central US`), as this is the recommendation to optimize the DNS resolution. They pay close attention to the replicated regions, as they want no more regions than needed since the business unit (bu0001) incurs Premium SKU registry fees for _each region_ they geo-replicate to. Under this configuration, they will pay *two* times Premium per month to get region proximity and to ensure no extra network egress fees from distant regions.
```

05-cluster-prerequisites.md
```diff
@@ -35,14 +35,25 @@ Following the steps below will result in the provisioning of the shared Azure re
...
+  >
+  > The app team is instructed to build the smallest container images they can. This is something they could achieve by following the [builder pattern](https://docs.docker.com/develop/develop-images/multistage-build/#before-multi-stage-builds) or the [mutli-stage builds](https://docs.docker.com/develop/develop-images/multistage-build/#use-multi-stage-builds), both approaches will produce final smaller container images that are meant to be for runtime-only. This will be beneficial in many ways but mainly in the speed of replication as well as in the transfer costs. A great feature as part of the `Geo-Replication` is that it will only replicates unique layers, also reducing data transfer across regions.
```
ckittel, line 50:
```suggestion
  > The app team is instructed to build the smallest container images they can. This is something they could achieve by following the [builder pattern](https://docs.docker.com/develop/develop-images/multistage-build/#before-multi-stage-builds) or the [mutli-stage builds](https://docs.docker.com/develop/develop-images/multistage-build/#use-multi-stage-builds). Both approaches will produce final smaller container images that are meant to be for runtime-only. This will be beneficial in many ways but mainly in the speed of replication as well as in the transfer costs. A key feature as part of ACR's geo-replication is that it will only replicate unique layers, also further reducing data transfer across regions.
```

05-cluster-prerequisites.md
```diff
@@ -35,14 +35,25 @@ Following the steps below will result in the provisioning of the shared Azure re
...
+  >
+  > In case of a region is down, the app team is now covered by the Azure Traffic Manager that comes on the scene to help deriving traffic to the registry located in the region that is closest to their multiple clusters in terms of network latency.
```
ckittel, line 52: I assume ATM is an implementation detail of ACR?  Or is that a new addition to the deployment here?  If it's an implementation detail of ACR, you might want to reword this so customers aren't confused if they don't "see" an ATM in the deployment.

05-cluster-prerequisites.md
```diff
@@ -35,14 +35,25 @@ Following the steps below will result in the provisioning of the shared Azure re
...
+  >
+  > After this initial design decision at the ACR level, the app team can also consider analyzing how they could tactically expand into [Availability Zones](https://docs.microsoft.com/en-us/azure/container-registry/zone-redundancy) as a way of being even more resilient.
```
ckittel, line 54:
```suggestion
  > After this initial design decision at the ACR level, the app team can also consider analyzing how they could tactically expand into [Availability Zones](https://docs.microsoft.com/azure/container-registry/zone-redundancy) as a way of being even more resilient.
```

05-cluster-prerequisites.md
```diff
@@ -35,14 +35,25 @@ Following the steps below will result in the provisioning of the shared Azure re
...
+  >
+  > :bulb: Another benefit of having `Geo-Replication` is that security permissions are now centralized in a single registry instance. That said, if your organization shares the _User Identities_ among the container hosts, this simplifies the security management a lot. In this reference implementation, the design decision is not to share a single _User Identity_ among the multiple clusters, otherwise that single identity will be granted with permissions to access other Azure resources that are meant to be cluster-specific such us Azure Key Vault. As things develop, the combination of [Availability Zones](https://docs.microsoft.com/en-us/azure/container-registry/zone-redundancy), currently in _Preview_,  for redundancy within a region, and geo-replication across multiple regions, is the recommendation when looking for the highest reliability and performance of a registry.
```
ckittel, line 56: > That said, if your organization shares the _User Identities_ among the container hosts, this simplifies the security management a lot.

Maybe just remove this bit, since kubelet will get its own identity in each cluster by default (and by implementation restrictions on the service currently).  The edge case where you're using pull secrets with logins probably doesn't need to be addressed here.  But open to more convo around this if you want.

ckittel, line 56:
```suggestion
  > :bulb: Another benefit of having `Geo-Replication` is that security permissions are now centralized in a single registry instance. That said, if your organization shares the _User Identities_ among the container hosts, this simplifies the security management a lot. In this reference implementation, the design decision is not to share a single _User Identity_ among the multiple clusters, otherwise that single identity will be granted with permissions to access other Azure resources that are meant to be cluster-specific such us Azure Key Vault. As things develop, the combination of [Availability Zones](https://docs.microsoft.com/azure/container-registry/zone-redundancy), currently in _Preview_,  for redundancy within a region, and geo-replication across multiple regions, is the recommendation when looking for the highest reliability and performance of a container registry.
```

ckittel (approved):
Couple things inline + what we talked about on Teams.  Feel free to adjust/take/leave what you wish.  🚀

05-cluster-prerequisites.md
ferantivero: done | Addressed ^ from [d0f97](https://github.com/mspnp/aks-secure-baseline/pull/170/commits/d0f9791107b40902cbfba5d895ee9a198adf49c4)

Additionally, [an additional request for re-wording](https://github.com/mspnp/aks-secure-baseline/pull/170#discussion_r585728106) is now address from [here](https://github.com/mspnp/aks-secure-baseline/pull/170/commits/0bc71fc8e6800f2c8c5764fb2109aba272b577d8). Originally resolved because it was delivered as a bundle with a suggested change that had been accepted.

05-cluster-prerequisites.md
ferantivero: done | addressed from [67c7e](https://github.com/mspnp/aks-secure-baseline/pull/170/commits/67c7ecce45c3b3aacca1d054fc2b80ff2d6c2aed)

05-cluster-prerequisites.md
ferantivero: oh yes, btw ATM is an implementation detail of ACR.

ckittel (approved):

PR #180, opened 2021-03-09
[Multi Cluster] [Management at Scale] Centralized Log Analytics validation

ferantivero:
- modify previous validation to be less workload specific and consider the multi cluster infrastructure.
- add some narratives around Traffic Analytics
- add reminder on having some buffer in your network segmentation to support extra IP allocations under a Failover event


03-networking.md
```diff
@@ -105,6 +105,14 @@ The following two resource groups will be created and populated with networking
...
 
+## Preparing for a Failover
+
+> :book: the networking team is now dealing with multiple clusters in different regions. Understanding how the traffic flows at layers 4 and 7 through their deployed networking topology is now more critical than ever. That's why the team is looking at the different tooling that could provide monitoring over their networks.  One of the Azure Monitor products at subscription level is Network Watcher that offers two really interesting features such us NSG Flow Logs and with that [Traffic Analytics](https://docs.microsoft.com/en-us/azure/network-watcher/traffic-analytics). The latter can bring some light over the table when it is about analyzing traffic like from where it is being originated, how is flowing thought the different regions, or how much is bening vs malicious together with many more details at the security and performance level. [With no upfront cost and no termination fees](https://azure.microsoft.com/en-us/pricing/details/network-watcher/) the business unit (BU0001) is charged for collection and processing logs per GB at 10-min or 60-min intervals.
```
ckittel, line 110:
```suggestion
> :book: The networking team is now dealing with multiple clusters in different regions. Understanding how the traffic flows at layers 4 and 7 through their deployed networking topology is now more critical than ever. That's why the team is looking at the different tooling that could provide monitoring over their networks.  One of the Azure Monitor products at subscription level is Network Watcher that offers two really interesting features such as NSG Flow Logs and with that [Traffic Analytics](https://docs.microsoft.com/azure/network-watcher/traffic-analytics). The latter can bring some light over the table when it is about analyzing traffic like from where it is being originated, how it is flowing thought the different regions, or how much is benign vs malicious together with many more details at the security and performance level. [With no upfront cost and no termination fees](https://azure.microsoft.com/pricing/details/network-watcher/) the business unit (BU0001) is charged for collection and processing logs per GB at 10-min or 60-min intervals.
```

03-networking.md
```diff
@@ -105,6 +105,14 @@ The following two resource groups will be created and populated with networking
...
+
+![Traffic Analytics Geo Map View of the AKS Multi Cluster reference implementation under load. Traffic is coming from a single Azure Front Door POP and is distrubuted to both regions after the first failover is complete](images/traffic-analytics-geo-map.png)
+
+> :bulb: The [AKS Baseline](https://github.com/mspnp/aks-secure-baseline) has already covered the how(s) and why(s) of the current [network topology segmentation](https://github.com/mspnp/aks-secure-baseline/blob/main/networking/topology.md). But something that is worth to remember while preparing for a high availability archicture is that the network needs to be right sized to absorb a sudden increase in traffic that might request twice the number of IPs when scheduling more _Pods_ to failover a region.
```
ckittel, line 114:
```suggestion
> :bulb: The [AKS Baseline](https://github.com/mspnp/aks-secure-baseline) has already covered the how(s) and why(s) of the current [network topology segmentation](https://github.com/mspnp/aks-secure-baseline/blob/main/networking/topology.md). But something that is worth to remember while preparing for a high availability architecture is that the network needs to be right sized to absorb a sudden increase in traffic that might request twice the number of IPs when scheduling more _Pods_ to failover a region.
```

ckittel (approved):
w00t.  Reads good.  Only really minor typos below.

PR #185, opened 2021-03-25
[Multi Cluster] Canary Testing

ferantivero:
just incorporate the narratives and bits to support the canary testing story this multi cluster/regions could potentially enable.


09-workload.md
```diff
@@ -6,37 +6,48 @@ The cluster now has an [Traefik configured with a TLS certificate](./09-secret-m
...
+   kubectl apply -f ./workload/aspnetapp.yaml --context $AKS_CLUSTER_NAME_BU0001A0042_03
    ```
 
-1. Wait until is ready to process requests running
+1. Deploy the same workload as another instance in the second AKS cluster deployed to Region 2. Additionally, you will be enable some canary testing from the second cluster.
+
+   > :book: The app team is now more confident than ever with its second regional AKS cluster. The team follows an active/active High Availability strategy, and it is known that the major stream of clients comes from East US 2. Therefore, they realize they have some iddle resources most of the time from Central US. This makes them to consider starting some canary testing in that specific region as this will be introducing a very low risk for the normal operatory. Using a weighted load balanced strategy, clients will be attended most of the time by their well known and stable ASP.NET Core 3.1 workload app, while others by new ASP.NET 5 aleatory. From this experimentation, the app team wants to evaluate the memory utlization of their workload using the new ASP.NET version 5. This will allow them to plan ahead before a full migration.
```
ckittel, line 19:
```suggestion
   > :book: The app team is now more confident than ever with its second regional AKS cluster. The team follows an active/active High Availability strategy, and it is known that the major stream of clients comes from East US 2. Therefore, they realize they have some idle resources most of the time from Central US. This makes them to consider starting some canary testing in that specific region as this will be introducing a very low risk for normal operations. Using a weighted load balanced strategy, clients will be attended most of the time by their well known and stable ASP.NET Core 3.1 workload app, while others by new ASP.NET 5 aleatory. From this experimentation, the app team wants to evaluate the memory utilization of their workload using the new ASP.NET version 5. This will allow them to plan ahead before a full migration.
```

09-workload.md
```diff
@@ -6,37 +6,48 @@ The cluster now has an [Traefik configured with a TLS certificate](./09-secret-m
...
+
+   > :warning: Please note that this canary testing story is not recommended for organizations that are operating critical productive systems. But this is a good example of what are the options being enabled when adding more availability to your application, and having some iddle resources to be employed with care.
```
ckittel, line 21:
```suggestion
   > :warning: Please note that this canary testing story is not recommended for organizations that are operating critical production systems as peak utilization. But this is a good example of what are the options being enabled when adding more availability to your application, and having some idle resources to be employed with care.
```

workload/aspnetapp-canary.yaml
```diff
@@ -0,0 +1,133 @@
...
+                  values:
+                  - traefik-ingress-ilb
+              topologyKey: "kubernetes.io/hostname"
+      containers:
+      - name: aspnet-webapp-sample
+        image: mcr.microsoft.com/dotnet/samples:aspnetapp
+        imagePullPolicy: Always
+        securityContext:
+          allowPrivilegeEscalation: false
+          readOnlyRootFilesystem: true
```
ckittel, line 54: I was reading that this is literally unsupported for aspnet images.  Even with the COMPlus_EnableDiagnotics flag set. ASP.NET will page to filesystem if it needs to.  Sounds like it could be dangerous to block this from happening.  I know we are not here to show "the best possible aspnetcore" container spec ... but just thought I would mention it.  I caught it in regulated and remove this (or set it to false, I forget which)

ckittel: Also, I know completely unrealted to what your change is.  Take/leave/backlog -- as  you wish.

workload/aspnetapp.yaml
```diff
@@ -95,40 +95,33 @@ spec:
     port: 80
     targetPort: 8080
 ---
-apiVersion: networking.k8s.io/v1
-kind: Ingress
+apiVersion: traefik.containo.us/v1alpha1
```
ckittel, line 98: Ahh, moving to CRD vs native Ingress.  I like the fact that the baseline doesn't, but this does.  That shows, once again, how you materalize your cluster is really based on your NEEDS, but it can still all be founded on the baseline. (You can auto-resolve this)

workload/aspnetapp.yaml
```diff
@@ -95,40 +95,33 @@ spec:
...
+      port: 80
+      namespace: a0042
+      kind: Service
+    middlewares:
+      - name: app-gateway-snet@file
+      - name: gzip-compress@file
+  # it is possible to opt for certificate management strategy with dedicated
+  # certificates for each TLS SNI route.
+  # In this Rereference Implementation for the sake of simplicity we use a
+  # wildcard default certificate added at Ingress Controller configuration level which is *.example.com
```
ckittel, line 125:
```suggestion
  # wildcard default certificate added at Ingress Controller configuration level which is *.aks-ingress.contoso.com
```

I know this exists this way in the baseline as well, but maybe we can change it here.

09-workload.md
```diff
@@ -6,37 +6,48 @@ The cluster now has an [Traefik configured with a TLS certificate](./09-secret-m
...
 
-1. Deploy the same workload as another instance in the second AKS cluster
+> :book: The app team is happy to confirm to the BU0001 that their workload is now consuming less memory thanks to this update.
 
-   ```bash
-   kubectl create -f ./workload/aspnetapp.yaml --context $AKS_CLUSTER_NAME_BU0001A0042_04
-   kubectl wait -n a0042 --for=condition=ready pod --selector=app.kubernetes.io/name=aspnetapp --timeout=90s --context $AKS_CLUSTER_NAME_BU0001A0042_04
-   kubectl get ingress aspnetapp-ingress -n a0042 --context $AKS_CLUSTER_NAME_BU0001A0042_04
-   ```
+![The app team is doing some canary testing of their workload, just a few requests are routed to the new version. The ASP.NET 5 workload app is reporting less memory usage.](images/canary-testing.gif)
```
ckittel, line 50: I like it! Do we need any "now remove the 3.1 version?"  Or we okay leaving it in this state? If we leave it in this state, maybe we can at least call out that we're just leaving it like this, but of course the app team would have migrated to the new version after validation.

ckittel (approved):
Couple suggestions inline.

09-workload.md
ferantivero: at this point we have:

region 1 - ASP.NET Core 3.1 (100%)
region 2 - ASP.NET Core 3.1 (75%) - ASP.NET 5 (25%)

how do you think we should proceed from narratives perspective. I was thinking to keep this "as-is"

workload/aspnetapp.yaml
ferantivero: <3

workload/aspnetapp-canary.yaml
ferantivero: yeah, the thing is that we got reports with some warnings at least when configuring it the other way around.

changed | Addressed [69318](https://github.com/mspnp/aks-secure-baseline/pull/185/commits/693185325032f8e433a591bff08a3d447d79b2d0)

09-workload.md
ferantivero: Just created a AzDO item to follow up on this, so we can keep discussion around.

PR #187, opened 2021-03-30
Update to AKS 1.20

v-fearam:
- CSI driver and provider
- ASP net app
- AKS 1.20

### DIFF
--- a/inner-loop-scripts/shell/1-cluster-stamp.sh
+++ b/inner-loop-scripts/shell/1-cluster-stamp.sh
@@ -72,31 +72,13 @@ kubectl create namespace cluster-baseline-settings
 kubectl apply -f ../../cluster-manifests/cluster-baseline-settings/flux.yaml
 kubectl wait --namespace cluster-baseline-settings --for=condition=ready pod --selector=app.kubernetes.io/name=flux --timeout=90s
 
+ACR_NAME=$(az deployment group show -g $RGNAMECLUSTER -n cluster-0001 --query properties.outputs.containerRegistryName.value -o tsv)
+# Import ingress controller image hosted in public container registries
+az acr import --source docker.io/library/traefik:2.2.1 -n $ACR_NAME
+
 echo ""
 echo "# Creating AAD Groups and users for the created cluster"
 echo ""
-
-# We are going to use a the new tenant which manage the cluster identity
-az login --allow-no-subscriptions -t $TENANT_ID
-
-#Creating AAD groups which will be associated to k8s out of the box cluster roles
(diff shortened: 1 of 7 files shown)


inner-loop-scripts/shell/1-cluster-stamp.sh
```diff
@@ -72,32 +72,14 @@ kubectl create namespace cluster-baseline-settings
 kubectl apply -f ../../cluster-manifests/cluster-baseline-settings/flux.yaml
 kubectl wait --namespace cluster-baseline-settings --for=condition=ready pod --selector=app.kubernetes.io/name=flux --timeout=90s
 
+ACR_NAME=$(az deployment group show -g $RGNAMECLUSTER -n cluster-0001 --query properties.outputs.containerRegistryName.value -o tsv)
+# Import ingress controller image hosted in public container registries
+az acr import --source docker.io/library/traefik:2.2.1 -n $ACR_NAME
```
ckittel, line 77: We need to do this in the instructions as well, right?

workload/aspnetapp.yaml
```diff
@@ -46,8 +46,8 @@ spec:
                   - traefik-ingress-ilb
               topologyKey: "kubernetes.io/hostname"
       containers:
-      - name: aspnetcore-webapp-sample
-        image: mcr.microsoft.com/dotnet/core/samples:aspnetapp
+      - name: aspnet-webapp-sample
+        image: mcr.microsoft.com/dotnet/samples:aspnetapp
         imagePullPolicy: Always
         securityContext:
           allowPrivilegeEscalation: false
```
ckittel, line 53: Not this line, but the line below (GH won't let me target it), I'm pretty sure is UNSUPPORTED for aspnet app files. Disk access is required for paging.  I know it's not in scope, but technically this change wasn't either. I think you can set this to `false` and then remove the `COMPlus_EnableDiagnostics` "workaround" below with the env variable.  I don't think that's the best practice for these base images.

ckittel (changes requested):
Only one question, and one other small request.  If you have any questions, just lmk!

inner-loop-scripts/shell/1-cluster-stamp.sh
v-fearam: No, it is there and it was not here
https://github.com/mspnp/aks-secure-baseline/blob/main/08-secret-managment-and-ingress-controller.md
point 5.

ckittel: Ah, got it.

workload/aspnetapp.yaml
v-fearam: I need to rebase, because the new Net5 app is already in master.
About your comment I need to investigate. Good point.

PR #188, opened 2021-03-30
Validated that free subscriptions are not possible.

ckittel:
Subscription used cannot be a free subscription due to significant quota (core and SKU) limitations.


ckittel:
Once this merges, I'll bubble to regulated and multi-region.

v-fearam (approved):

PR #189, opened 2021-03-31
Updating Cluster Version

v-fearam:
This PR is against update-csi-and-workload because this depends on it.


v-fearam:
I forgot about api version. 
I will test and include here.
"2021-02-01"
https://docs.microsoft.com/en-us/azure/templates/microsoft.containerservice/2021-02-01/managedclusters?tabs=json

PR #190, opened 2021-04-05
Updating Treafik Ingress Controller

v-fearam:
Updating treafik version (https://doc.traefik.io/traefik/v2.4/migration/v2/#ingressclass)
Updating Managed Cluster API Version

### DIFF
--- a/inner-loop-scripts/shell/1-cluster-stamp.sh
+++ b/inner-loop-scripts/shell/1-cluster-stamp.sh
@@ -74,7 +74,7 @@ kubectl wait --namespace cluster-baseline-settings --for=condition=ready pod --s
 
 ACR_NAME=$(az deployment group show -g $RGNAMECLUSTER -n cluster-0001 --query properties.outputs.containerRegistryName.value -o tsv)
 # Import ingress controller image hosted in public container registries
-az acr import --source docker.io/library/traefik:2.2.1 -n $ACR_NAME
+az acr import --source docker.io/library/traefik:v2.4.8 -n $ACR_NAME
 
 echo ""
 echo "# Creating AAD Groups and users for the created cluster"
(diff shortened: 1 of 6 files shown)


inner-loop-scripts/shell/1-cluster-stamp.sh
```diff
@@ -74,7 +74,7 @@ kubectl wait --namespace cluster-baseline-settings --for=condition=ready pod --s
 
 ACR_NAME=$(az deployment group show -g $RGNAMECLUSTER -n cluster-0001 --query properties.outputs.containerRegistryName.value -o tsv)
 # Import ingress controller image hosted in public container registries
-az acr import --source docker.io/library/traefik:2.2.1 -n $ACR_NAME
+az acr import --source docker.io/library/traefik:v2.4.8 -n $ACR_NAME
```
ckittel, line 77: YOu need to do this in `08-secret-managment-and-ingress-controller.md` as well, right?

ckittel (changes requested):
I didn't test end to end, assume you did.  Only one obersvation that will prevent folks from completing this if they are using the readme files.

inner-loop-scripts/shell/1-cluster-stamp.sh
v-fearam: Yes. Good catch. Thanks. Fixed

v-fearam:
yes. I tested end to end before creating the pull request.

inner-loop-scripts/shell/1-cluster-stamp.sh
ckittel: coo!

ckittel (approved):
🚢

PR #194, opened 2021-04-21
Adding Azure Firewall Manager and Azure Firewall Policies

v-fearam:
Also was updated the AKS version to 1.20.5
It was added a base Firewall Policy

### DIFF
--- a/networking/hub-default.json
+++ b/networking/hub-default.json
@@ -364,8 +364,100 @@
             }
         },
         {
+            "type": "Microsoft.Network/firewallPolicies",
+            "apiVersion": "2020-11-01",
+            "name": "[variables('hubFwName')]",
+            "location": "[parameters('location')]",
+            "properties": {
+                "sku": {
+                    "tier": "Standard"
+                },
+                "threatIntelMode": "Deny",
+                "threatIntelWhitelist": {
+                    "ipAddresses": []
+                },
+                "dnsSettings": {
+                    "servers": [],
+                    "enableProxy": true
+                }
+            },
+            "resources": [
+                {
+                    "type": "Microsoft.Network/firewallPolicies/ruleCollectionGroups",
+                    "apiVersion": "2020-11-01",
+                    "name": "[concat(variables('hubFwName'), '/DefaultDnatRuleCollectionGroup')]",
+                    "location": "[parameters('location')]",
+                    "dependsOn": [
+                        "[variables('hubFwName')]"
+                    ],
+                    "properties": {
+                        "priority": 100,
+                        "ruleCollections": []
+                    }
+                },
+                {
+                    "type": "Microsoft.Network/firewallPolicies/ruleCollectionGroups",
+                    "apiVersion": "2020-11-01",
+                    "name": "[concat(variables('hubFwName'), '/DefaultApplicationRuleCollectionGroup')]",
+                    "location":"[parameters('location')]",
+                    "dependsOn": [
+                        "[variables('hubFwName')]",
+                        "[resourceId('Microsoft.Network/firewallPolicies/ruleCollectionGroups', variables('hubFwName'), 'DefaultDnatRuleCollectionGroup')]"
+                    ],
+                    "properties": {
+                        "priority": 300,
+                        "ruleCollections": []
+                    }
+                },
+                {
+                    "type": "Microsoft.Network/firewallPolicies/ruleCollectionGroups",
+                    "apiVersion": "2020-11-01",
+                    "name": "[concat(variables('hubFwName'), '/DefaultNetworkRuleCollectionGroup')]",
+                    "location": "[parameters('location')]",
+                    "dependsOn": [
+                        "[variables('hubFwName')]",
+                        "[resourceId('Microsoft.Network/firewallPolicies/ruleCollectionGroups', variables('hubFwName'), 'DefaultApplicationRuleCollectionGroup')]"
+                    ],
+                    "properties": {
+                        "priority": 200,
+                        "ruleCollections": [
+                            {
+                                "ruleCollectionType": "FirewallPolicyFilterRuleCollection",
+                                "action": {
+                                    "type": "Allow"
+                                },
+                                "rules": [
+                                    {
+                                        "ruleType": "NetworkRule",
+                                        "name": "DNS",
+                                        "ipProtocols": [
+                                            "UDP"
+                                        ],
+                                        "sourceAddresses": [
+                                            "*"
+                                        ],
+                                        "sourceIpGroups": [],
+                                        "destinationAddresses": [
+                                            "*"
+                                        ],
+                                        "destinationIpGroups": [],
+                                        "destinationFqdns": [],
+                                        "destinationPorts": [
+                                            "53"
+                                        ]
+                                    }
+                                ],
+                                "name": "org-wide-allowed",
+                                "priority": 100
+                            }
+                        ]
+                    }
+                }
+          ]
+        },
+        {
             "type": "Microsoft.Network/azureFirewalls",
-            "apiVersion": "2020-05-01",
+            "apiVersion": "2020-11-01",
             "name": "[variables('hubFwName')]",
             "location": "[parameters('location')]",
             "comments": "This is the regional Azure Firewall that all regional spoke networks can egress through.",
@@ -376,12 +468,11 @@
             ],
             "dependsOn": [
                 "create-fw-pips",
-                "[resourceId('Microsoft.Network/virtualNetworks', variables('hubVnetName'))]"
+                "[resourceId('Microsoft.Network/virtualNetworks', variables('hubVnetName'))]",
+                "[resourceId('Microsoft.Network/firewallPolicies', variables('hubFwName'))]"
             ],
             "properties": {
-                "additionalProperties": {
-                    "Network.DNS.EnableProxy": "true"
-                },
+                "additionalProperties": {},
                 "sku": {
                     "name": "AZFW_VNet",
                     "tier": "Standard"
@@ -417,36 +508,11 @@
                     }
                 ],
                 "natRuleCollections": [],
-                "networkRuleCollections": [
-                    {
-                        "name": "org-wide-allowed",
-                        "properties": {
-                            "action": {
-                                "type": "Allow"
-                            },
-                            "priority": 100,
-                            "rules": [
-                                {
-                                    "name": "DNS",
-                                    "description": "Consider restricting this to only DNS servers you expect to be used by spokes of this hub.",
-                                    "sourceAddresses": [
-                                        "*"
-                                    ],
-                                    "protocols": [
-                                        "UDP"
-                                    ],
-                                    "destinationPorts": [
-                                        "53"
-                                    ],
-                                    "destinationAddresses": [
-                                        "*"
-                                    ]
-                                }
-                            ]
-                        }
-                    }
-                ],
-                "applicationRuleCollections": []
+                "networkRuleCollections": [],
+                "applicationRuleCollections": [],
+                "firewallPolicy": {
+                    "id": "[resourceId('Microsoft.Network/firewallPolicies', variables('hubFwName'))]"
+                }
             },
             "resources": [
                 {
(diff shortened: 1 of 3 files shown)


networking/hub-regionA.json
```diff
@@ -387,9 +387,464 @@
...
+                        ]
+                    }
+                },
+                {
+                    "type": "Microsoft.Network/firewallPolicies/ruleCollectionGroups",
+                    "apiVersion": "2020-11-01",
+                    "name": "[concat(variables('hubFwName'), '/DefaultNetworkRuleCollectionGroup')]",
+                    "location": "[parameters('location')]",
+                    "dependsOn": [
+                        "[variables('hubFwName')]",
```
ckittel, line 719: Is this supposed to be the policy, if so, can we reference it more specifically.

networking/hub-regionA.json
```diff
@@ -387,9 +387,464 @@
...
+                        "ruleCollections": []
+                    }
+                },
+                {
+                    "type": "Microsoft.Network/firewallPolicies/ruleCollectionGroups",
+                    "apiVersion": "2020-11-01",
+                    "name": "[concat(variables('hubFwName'), '/DefaultApplicationRuleCollectionGroup')]",
+                    "location":"[parameters('location')]",
+                    "dependsOn": [
+                        "[variables('hubFwName')]",
```
ckittel, line 431: Is this supposed to be the policy, if so, can we reference it more specifically.

networking/hub-default.json
```diff
@@ -363,9 +363,101 @@
...
+                        "ruleCollections": []
+                    }
+                },
+                {
+                    "type": "Microsoft.Network/firewallPolicies/ruleCollectionGroups",
+                    "apiVersion": "2020-11-01",
+                    "name": "[concat(variables('hubFwName'), '/DefaultApplicationRuleCollectionGroup')]",
+                    "location":"[parameters('location')]",
+                    "dependsOn": [
+                        "[variables('hubFwName')]",
```
ckittel, line 404: Can we be more specific about this depends?

networking/hub-default.json
```diff
@@ -363,9 +363,101 @@
...
+                        "[resourceId('Microsoft.Network/firewallPolicies/ruleCollectionGroups', variables('hubFwName'), 'DefaultDnatRuleCollectionGroup')]"
```
ckittel, line 405: This depends on itself?

networking/hub-default.json
```diff
@@ -363,9 +363,101 @@
...
+                    }
+                },
+                {
+                    "type": "Microsoft.Network/firewallPolicies/ruleCollectionGroups",
+                    "apiVersion": "2020-11-01",
+                    "name": "[concat(variables('hubFwName'), '/DefaultNetworkRuleCollectionGroup')]",
+                    "location": "[parameters('location')]",
+                    "dependsOn": [
+                        "[variables('hubFwName')]",
+                        "[resourceId('Microsoft.Network/firewallPolicies/ruleCollectionGroups', variables('hubFwName'), 'DefaultApplicationRuleCollectionGroup')]"
```
ckittel, line 419: These are accurate?

ckittel (commented):
Just one common question, are the `dependsOn` right here?

networking/hub-default.json
v-fearam: It not it self. It is the same type, not the same resource.
You get a deploy error if you deploy all  Microsoft.Network/firewallPolicies/ruleCollectionGroups deploy together.
The error  was "there is another update operation executing, bad request"
The only solution solution was a dependency between them. It was the biggest issue of this task. 
The first ruleCollectionGroups  doesn't any dependency
the second depends on the first  ruleCollectionGroups 
the third depends on the second ruleCollectionGroups .

networking/hub-default.json
v-fearam: yes, it is needed to avoid deploy error. See the another comment

networking/hub-regionA.json
v-fearam: let me try it out. I will let you know.

ckittel:
Should we be building off of `basePolicy` though more?  I mean we want to make sure we're using this feature the way it was designed, to ensure we have easy-to-manage policies for the regional (or even multi) hubs

networking/hub-default.json
ckittel: Ah,  you're right, I read that poorly, closing this one comment.

networking/hub-regionA.json
ckittel, line 719: Doesn't this depend on the ipGroup?  Looks like ipGroup dependencies are not here?  Same for the others in here.

v-fearam: done

networking/hub-regionA.json
v-fearam: done

networking/hub-regionA.json
v-fearam: the dependency is on the root object, then the children are not needed to be added. 
You need to add on the children, if you define the ruleCollectionGroups outside the Microsoft.Network/firewallPolicies scope.

 "type": "Microsoft.Network/firewallPolicies",
            "apiVersion": "2020-11-01",
            "name": "[variables('hubFwName')]",
            "location": "[parameters('location')]",
            "dependsOn": [
                **"[resourceId('Microsoft.Network/ipGroups', variables('aksIpGroupName'))]"**
            ],

networking/hub-default.json
v-fearam: done

v-fearam:
About base policies
https://docs.microsoft.com/en-us/azure/firewall-manager/policy-overview#hierarchical-policies
It is not mandatory, the idea is,  if you have a basic organization policies you can define on the baseline policy and then inherit from it.
We need to invent that situation here, In my opinion is out of scope of the RI story

ckittel:
Don't we have basic org polices?  I think we even call them "ORG WIDE" or something.  Hence my question :)

networking/hub-regionA.json
```diff
@@ -415,7 +415,7 @@
                     "name": "[concat(variables('hubFwName'), '/DefaultDnatRuleCollectionGroup')]",
                     "location": "[parameters('location')]",
                     "dependsOn": [
-                        "[variables('hubFwName')]"
+                        "[concat('Microsoft.Network/firewallPolicies/', variables('hubFwName'))]"
```
ckittel, line 418: shouldn't use `concat` to do references, needs to be `resourceId` -- this will trip TTK and other validators  Can you reference this (and the others you changed) via the standard `resourceId` function please?

networking/hub-regionA.json
ckittel: That seems like that could cause a issue if someone breaks the dependency chain here, especially since the dependency chain is currently kinda a workaround and artificial.  Would it be wrong to add that here?

networking/hub-default.json
v-fearam: image
the error deleting the dependency 
{
    "status": "Failed",
    "error": {
        "code": "FirewallPolicyRuleCollectionGroupUpdateNotAllowed",
        "message": "Rule Collection Group DefaultDnatRuleCollectionGroup can not be updated because Parent Firewall Policy fw-eastus2 is in Updating state from previous operation ID : f21d4ad1-1db0-44d2-8c0e-ba2b0cfa8b15."
    }
}

v-fearam:
About "Don't we have basic org polices? I think we even call them "ORG WIDE" or something. Hence my question :) And then in the multi-region, you'll be able to inherit it twice."

I will add task for next sprint..... Maybe in multicluster we can show it better later on.

networking/hub-regionA.json
v-fearam: after adding
"[resourceId('Microsoft.Network/firewallPolicies', variables('hubFwName'))]",
I got the error 
  "message": "Put on Firewall Policy fw-eastus2 Failed with 1 faulted referenced firewalls"

The recomendation is the simpler name
https://docs.microsoft.com/en-us/azure/azure-resource-manager/templates/child-resource-name-type?tabs=json#within-parent-resource

networking/hub-regionA.json
v-fearam: Would it be wrong to add that to each one?
no, but I preferred  avoid redundancy. 
I could change if you prefer.

The artificial dependency among  ruleCollectionGroups is not playing here. 
Here the dependency playing is between the parent resource (Microsoft.Network/firewallPolicies) and its children resources (the collection of Microsoft.Network/firewallPolicies/ruleCollectionGroups)

ckittel: You're right about that, I should have looked closer when I commented, but that resource doesn't ACTUALLY depend on the ip group, so I think it would be more correct to only list it (even if duplicated) on the ones that to.

networking/hub-regionA.json
ckittel: Ah! maybe the right thing to do is to name the policy not the same name as the firewall, that'll help keep those two separate, otherwise you have artificially created a 1o1 mapping by name alone.  If you ever applied that to another firewall, it would be "weird" with that name.

networking/hub-default.json
ckittel: I'm not sure what this means (as far as this comment thread)

networking/hub-regionA.json
ckittel: Good point, I think the fact that it's the FW name really throws it off for me.  If this was the name of the firewallPolicy, that would help readability more.  Sorry for sending you down the wrong path here.

networking/hub-default.json
v-fearam: the error that you get deleting this comment dependency mentioned before 
```
The first ruleCollectionGroups doesn't any dependency
the second depends on the first ruleCollectionGroups
the third depends on the second ruleCollectionGroups .
```

networking/hub-regionA.json
v-fearam: ok, I will do it.

networking/hub-default.json
ckittel: deleting a comment dependency?  I'm not sure I follow.  Sorry.  Feel free to ping me on teams, I must be missing something.

networking/hub-regionA.json
v-fearam: Agree, I will change the name. I will create a new name for the policy

networking/hub-default.json
ckittel: OH, I get it now. I misread something you wrote -- interpreted it another way.  Marking this resolved.

networking/hub-regionA.json
v-fearam: it was moved down

networking/hub-regionA.json
v-fearam: I will do on Monday

ckittel: Sounds good.  Have a wonderful weekend @v-fearam

v-fearam: done

networking/hub-default.json
```diff
@@ -70,6 +70,8 @@
             "[concat(variables('baseFwPipName'), '-02')]"
         ],
         "hubFwName": "[concat('fw-', parameters('location'))]",
+        "fwPoliciesBaseName": "[concat('fw-policies-base')]",
```
ckittel, line 73: was something supposed to be `concat`'d here?

networking/hub-regionA.json
```diff
@@ -78,6 +78,8 @@
             "[concat(variables('baseFwPipName'), '-02')]"
         ],
         "hubFwName": "[concat('fw-', parameters('location'))]",
+        "fwPoliciesBaseName": "[concat('fw-policies-base')]",
```
ckittel, line 81: Just a reminder to update it here based on the result of the other comment.

ckittel (changes requested):
Only one small question, otherwise, looks nice.  Helps to have a dedciated name for this.

networking/hub-default.json
v-fearam: my bad, I deleted the 'location' , I didn't delete the concat. Then in multicluster, this will be a general base policy on the 'share' resource group.
Good catch. I will fix it.

v-fearam: done

ckittel (approved):
🚀  Thanks!

PR #199, opened 2021-05-17
AzureAppGatewayAnalytics and fixing Firewall issue

v-fearam:
AzureAppGatewayAnalytics  was not working.
Fixing Firewall rules
Deleting useless firewall rules
Adding a dependency to avoid random issue

### DIFF
--- a/networking/hub-default.json
+++ b/networking/hub-default.json
@@ -70,7 +70,7 @@
             "[concat(variables('baseFwPipName'), '-02')]"
         ],
         "hubFwName": "[concat('fw-', parameters('location'))]",
-        "fwPoliciesBaseName": "fw-policies-base",
+        "fwPoliciesBaseName": "[concat('fw-policies-base-', parameters('location'))]", 
         "fwPoliciesName": "[concat('fw-policies-', parameters('location'))]",
         "hubVNetName": "[concat('vnet-', parameters('location'), '-hub')]",
         "bastionNetworkNsgName": "[concat('nsg-', parameters('location'), '-bastion')]",
--- a/networking/hub-default.json
+++ b/networking/hub-default.json
@@ -70,7 +70,7 @@
             "[concat(variables('baseFwPipName'), '-02')]"
         ],
         "hubFwName": "[concat('fw-', parameters('location'))]",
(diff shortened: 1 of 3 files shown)


networking/hub-default.json
```diff
@@ -70,7 +70,7 @@
             "[concat(variables('baseFwPipName'), '-02')]"
         ],
         "hubFwName": "[concat('fw-', parameters('location'))]",
-        "fwPoliciesBaseName": "fw-policies-base",
+        "fwPoliciesBaseName": "[concat('fw-policies-base-', parameters('location'))]",
```
ckittel, line 73: While policies are regional in where they need to be defined, it is my understanding that they can (and are designed to be) applied globally (meaning, to a FW in any region).  Base policies are likely non-regional in nature because they are designed to be applied globally at the enterprise.  What problem does adding region solve to the base policy?

ckittel (changes requested):
All good except for the open question I have on base policy name now including a region name

networking/hub-default.json
v-fearam: deleted.  It is without the region in the name, now.

networking/hub-default.json
```diff
@@ -70,7 +70,7 @@
             "[concat(variables('baseFwPipName'), '-02')]"
         ],
         "hubFwName": "[concat('fw-', parameters('location'))]",
-        "fwPoliciesBaseName": "fw-policies-base",
+        "fwPoliciesBaseName": "fw-policies-base",
```
ckittel, line 73:
```suggestion
        "fwPoliciesBaseName": "fw-policies-base",
```

networking/hub-regionA.json
```diff
@@ -78,7 +78,7 @@
             "[concat(variables('baseFwPipName'), '-02')]"
         ],
         "hubFwName": "[concat('fw-', parameters('location'))]",
-        "fwPoliciesBaseName": "fw-policies-base",
+        "fwPoliciesBaseName": "fw-policies-base",
```
ckittel, line 81:
```suggestion
        "fwPoliciesBaseName": "fw-policies-base",
```

ckittel (approved):
🚀

PR #207, opened 2021-07-01
Modified cluster stamp deployment file so it turns --admin switch off

magrande:


06-gitops.md
```diff
@@ -26,7 +26,7 @@ GitOps allows a team to author Kubernetes manifest files, persist them in their
...
...ration with your AKS cluster, a subscription Owner and Contributor or anyone who is a member of the Azure Kubernetes Service Cluster Admin Role can still access you AKS cluster by using the --admin switch; that isn't what we want to happen, so in order to prevent that practice, the cluster-stamp deployment file used in the [AKS Cluster creation step](05-aks-clusted.md), sets the property named 'disableLocalAccounts' to true which switches -–admin access off by default. In a following step, you'll log in with a user that has been added to the Azure AD security group used to back the Kubernetes RBAC admin role. Executing the first `kubectl` command below will invoke the AAD login process to auth the _user of your choice_, which will then be checked against Kubernetes RBAC to perform the action. The user you choose to log in with _must be a member of the AAD group bound_ to the `cluster-admin` ClusterRole. For simplicity you could either use the "break-glass" admin user created in [Azure Active Directory Integration](03-aad.md) (`bu0001a0008-admin`) or any user you assigned to the `cluster-admin` group assignment in your [`cluster-rbac.yaml`](cluster-manifests/cluster-rbac.yaml) file.
```
ckittel, line 29:
```suggestion
   > In the [Azure Active Directory Integration](03-aad.md) step, we placed our cluster under AAD group-backed RBAC. This is the first time we are seeing this used. `az aks get-credentials` sets your `kubectl` context so that you can issue commands against your cluster. Even when you have enabled Azure AD integration with your AKS cluster, an Azure user has sufficient permissions on the cluster resource can still access your AKS cluster by using the `--admin` switch to this command. Using this switch _bypasses_ Azure AD and uses client certificate authentication instead; that isn't what we want to happen. So in order to prevent that practice, local account access (such as cluster admin) is expressly disabled. In a following step, you'll log in with a user that has been added to the Azure AD security group used to back the Kubernetes RBAC admin role -- not as a Kubernetes local user account. Executing the first `kubectl` command below will invoke the AAD login process to auth the _user of your choice_, which will then be checked against Kubernetes RBAC to perform the action. The user you choose to log in with _must be a member of the AAD group bound_ to the `cluster-admin` ClusterRole. For simplicity you could either use the "break-glass" admin user created in [Azure Active Directory Integration](03-aad.md) (`bu0001a0008-admin`) or any user you assigned to the `cluster-admin` group assignment in your [`cluster-rbac.yaml`](cluster-manifests/cluster-rbac.yaml) file.
```

ckittel (changes requested):
Really good.  Only made a couple minor tweaks.

ckittel (approved):

PR #212, opened 2021-07-19
Add Event Grid System Topic for AKS

ckittel:
* Add Event Grid system topic for AKS
* Fix missing Azure RBAC step for  workflow deployment
* Fix deployment failure when admin and read-only group are the same
* resource property ordering
* Disable Local Accounts is GA


magrande (approved):
Approved ! Thanks for the detailed change list.

PR #214, opened 2021-07-30
Remove unused NTP network rule & CSE hackathon feedback

ckittel:
* Remove the firewall rule for NTP as chrony is now used (instead of ntpd) which gets time from hosts. Firewall rule no longer needed.
* Feedback from CSE hackathon
  * Use `resourceId()` instead of `concat()` in App GW
  * Updated to OWASP 3.2
  * Fixed poor naming of the 443 port in App GW
* "base 64" -> "Base64"

Completed an end-to-end deployment.


ferantivero (approved):
w00t; nice to see this get polished and amazing feedback @ckittel - LGTM 🚀

PR #218, opened 2021-08-17
Update deployment walkthrough to allow configurable domain

AAkindele:
Update the deployment walkthrough to allow "contoso.com" domain to be configurable.

- new environment variable and cluster-stamp.json ARM variable that default to "contoso.com"
- fixed sed command in the GitHub actions deployment option. missing pipes, "|", in the multi-line command
- added kustomization to allow host on the workload ingress to be updated
- tested using both option 1 and option 2 in the 05-aks-cluster.md instructions


09-workload.md
```diff
@@ -20,6 +20,49 @@ The cluster now has an [Traefik configured with a TLS certificate](./08-secret-m
...
+     tls:
+     - hosts:
+         - bu0001a0008-00.aks-ingress.$DOMAIN_NAME
+           # it is possible to opt for certificate management strategy with dedicated
+           # certificates for each TLS SNI route.
+           # In this Rereference Implementation for the sake of simplicity we use a
+           # wildcard default certificate added at Ingress Controller configuration level which is *.example.com
+           # secretName: <bu0001a0008-00-example-com-tls-secret>
+     rules:
+     - host: bu0001a0008-00.aks-ingress.$DOMAIN_NAME
```
ckittel, line 53:
```suggestion
     - host: bu0001a0008-00.aks-ingress.${DOMAIN_NAME}
```

09-workload.md
```diff
@@ -20,6 +20,49 @@ The cluster now has an [Traefik configured with a TLS certificate](./08-secret-m
...
+       kubernetes.io/ingress.class: traefik-internal
+       traefik.ingress.kubernetes.io/router.entrypoints: websecure
+       traefik.ingress.kubernetes.io/router.tls: "true"
+       traefik.ingress.kubernetes.io/router.tls.options: default
+       traefik.ingress.kubernetes.io/router.middlewares: app-gateway-snet@file, gzip-compress@file
+   spec:
+     # ingressClassName: "traefik-internal"
+     tls:
+     - hosts:
+         - bu0001a0008-00.aks-ingress.$DOMAIN_NAME
```
ckittel, line 46:
```suggestion
         - bu0001a0008-00.aks-ingress.${DOMAIN_NAME}
```

09-workload.md
```diff
@@ -20,6 +20,49 @@ The cluster now has an [Traefik configured with a TLS certificate](./08-secret-m
...
+         - path: /
+           pathType: Prefix
+           backend:
+             service:
+               name: aspnetapp-service
+               port:
+                 number: 80
+   EOF
+   ```
+
```
ckittel, line 65: We try to minimize inline yaml creation like this, (I know we have it in another spot here as well -- but trying to prevent more).  Any thoughts on maybe adding a Kustomization to the workload directory with a strategicMerge patch, then instead of this, we could do a quick `sed` command to replace a placeholder in that file, and instead of applying the workload with `-f` we could just `-k` that directory. There is a lot of "comment noise" in this file, which I think is best left in a file vs echo'd directly into a kubectl apply and to be lost.  Thoughts?

02-ca-certificates.md
```diff
@@ -30,7 +36,7 @@ Now that you have the [prerequisites](./01-prerequisites.md) met, follow the ste
    > :book: Contoso Bicycle will also procure another TLS certificate, a standard cert, to be used with the AKS Ingress Controller. This one is not EV, as it will not be user facing. Finally the app team decides to use a wildcard certificate of `*.aks-ingress.contoso.com` for the ingress controller.
 
    ```bash
-   openssl req -x509 -nodes -days 365 -newkey rsa:2048 -out traefik-ingress-internal-aks-ingress-contoso-com-tls.crt -keyout traefik-ingress-internal-aks-ingress-contoso-com-tls.key -subj "/CN=*.aks-ingress.contoso.com/O=Contoso Aks Ingress"
+   openssl req -x509 -nodes -days 365 -newkey rsa:2048 -out traefik-ingress-internal-aks-ingress-contoso-com-tls.crt -keyout traefik-ingress-internal-aks-ingress-contoso-com-tls.key -subj "/CN=*.aks-ingress.${DOMAIN_NAME}/O=Contoso Aks Ingress"
```
ckittel, line 39: Looks like the filenames here are going to be misleading as well.  I noticed that the Key Vault names suffer from the same thing (E.g. it's expecting the key to be called `traefik-ingress-internal-aks-ingress-contoso-com-tls`.  We should probably make those names more generic if we're going to go down this path, instead of it being mixed like it is being proposed.

09-workload.md
```diff
@@ -20,6 +20,49 @@ The cluster now has an [Traefik configured with a TLS certificate](./08-secret-m
...
+   spec:
+     # ingressClassName: "traefik-internal"
+     tls:
+     - hosts:
+         - bu0001a0008-00.aks-ingress.$DOMAIN_NAME
+           # it is possible to opt for certificate management strategy with dedicated
+           # certificates for each TLS SNI route.
+           # In this Rereference Implementation for the sake of simplicity we use a
+           # wildcard default certificate added at Ingress Controller configuration level which is *.example.com
+           # secretName: <bu0001a0008-00-example-com-tls-secret>
```
ckittel, line 51: This is set up as the default in Traefik's config, https://github.com/mspnp/aks-secure-baseline/blob/b3319376ac58fe28ad5244be2776bbf84633b47a/workload/traefik.yaml#L295 -- which has the other name hardcoded in there.  I'm thinking that secret name (KV and CSI KV Driver reference) should be genericized as well as part of this PR.

ckittel (changes requested):
Sorry for the delay @AAkindele -- took longer to unbury from my OOO than expected.  Feedback inline here.  If there is anything you have questions about, please don't hesitate to reach out to me.

09-workload.md
AAkindele: I like that approach. I'll work on making the changes.

02-ca-certificates.md
```diff
@@ -13,7 +19,7 @@ Now that you have the [prerequisites](./01-prerequisites.md) met, follow the ste
    Create the certificate for Azure Application Gateway with a common name of `bicycle.contoso.com`.
```
AAkindele, line 19: I'm struggling to come up with a good way to rephrase this. Maybe something like

Create the certificate for Azure Application Gateway with a common name for your domain. For example, `bicycle.contoso.com`."

09-workload.md
```diff
@@ -6,12 +6,20 @@ The cluster now has an [Traefik configured with a TLS certificate](./08-secret-m
 
 > :book: The Contoso app team is about to conclude this journey, but they need an app to test their new infrastructure. For this task they've picked out the venerable [ASP.NET Core Docker sample web app](https://github.com/dotnet/dotnet-docker/tree/master/samples/aspnetapp).
 
+1. Customize the host name of the Ingress resource to match your custom domain
+
+   > An example file is provided that will be used later with the built-in kubectl customization tool, Kustomize. This lets you patch the default Ingress resource definition with your own custom values.
+
+   ```bash
+   sed "s/<domain-name>/${DOMAIN_NAME}/g" workload/aspnetapp-ingress-patch.yaml.example > workload/aspnetapp-ingress-patch.yaml
```
ckittel, line 14: Part of a kustomization/workflow simplification suggestion (part 4 of 4):

1. Customize the host name of the Ingress resource to match your custom domain. _(Okay to skip this step if domain was left as contoso.com.)_

```suggestion
   sed -i "s/contoso.com/${DOMAIN_NAME}/" workload/aspnetapp-ingress-patch.yaml
```

09-workload.md
```diff
@@ -36,7 +44,8 @@ The cluster now has an [Traefik configured with a TLS certificate](./08-secret-m
    kubectl run curl -n a0008 -i --tty --rm --image=mcr.microsoft.com/azure-cli --limits='cpu=200m,memory=128Mi'
    
    # From within the open shell
-   curl -kI https://bu0001a0008-00.aks-ingress.contoso.com -w '%{remote_ip}\n'
+   DOMAIN_NAME="contoso.com"
```
ckittel, line 47:
```suggestion
   DOMAIN_NAME="contoso.com" # <-- Change to your custom domain value if a different one was used
```

09-workload.md
```diff
@@ -6,12 +6,20 @@ The cluster now has an [Traefik configured with a TLS certificate](./08-secret-m
 
 > :book: The Contoso app team is about to conclude this journey, but they need an app to test their new infrastructure. For this task they've picked out the venerable [ASP.NET Core Docker sample web app](https://github.com/dotnet/dotnet-docker/tree/master/samples/aspnetapp).
 
+1. Customize the host name of the Ingress resource to match your custom domain
+
+   > An example file is provided that will be used later with the built-in kubectl customization tool, Kustomize. This lets you patch the default Ingress resource definition with your own custom values.
+
```
ckittel, line 12: Part of a kustomization/workflow simplification suggestion (part 4 of 4):

```suggestion
1. Customize the host name of the Ingress resource to match your custom domain. _(You can skip this step if domain was left as contoso.com.)_

```

ckittel (changes requested):
Nice update.  It feels like we're on the right track here. I've got a series of suggestions here to simplify even more.  No need to have untracked file and duplicate all of the ingress configuration.  Can even make the patch update process optional (but safe to run even if they don't see that message).  What do ya think?

09-workload.md
ckittel: Looks like you've got them all updated, nice!

02-ca-certificates.md
ckittel: Looks like you found them all, nice.

09-workload.md
ckittel: Good start on this, I'm going to resolve this comment here, because my follow-on review has a slight tweak to the implementation, but the end result is the same.

02-ca-certificates.md
ckittel: Could even drop the common name part out:

```suggestion
   Create the certificate that will be presented to web clients by Azure Application Gateway for your domain.
```

What do ya think?

AAkindele: 👍 I like it.

AAkindele:
> Nice update. It feels like we're on the right track here. I've got a series of suggestions here to simplify even more. No need to have untracked file and duplicate all of the ingress configuration. Can even make the patch update process optional (but safe to run even if they don't see that message). What do ya think?

I like the suggestions. It reduces the ingress dup and still allows folks to customize. I'll add the changes, fix the merge conflict, and do another test run.

ckittel:
Sorry, just merged in @alfredoihernandez's PR on saving env variables -- which caused more merge conflicts.  I'll still review today.

ckittel (approved):
w00t, excellent @AAkindele -- I know there is an outstanding merge conflict (always a race condition :)), but as far as the actual changes here, they're spot on.  Thank you!

After you resolve the merge conflict, I'll go ahead and apply this PR.  <3

PR #219, opened 2021-08-18
move pdb to v1 now that it's GA in 1.21

ckittel:
1.21 PodDisruptionBudgets are GA, spec hasn't changed, so remove the beta from the API version to reduce a warning


magrande (commented):
Looks good. Approved

magrande (approved):

PR #222, opened 2021-08-31
Support MSDN and AAD guest accounts better

ckittel:
The AKV step sourced the domain information in such a way that it only worked for org users, but didn't work for org guests nor typical MSDN ("EXT") users.  Change how that is done to make it easier for users to get past that step.

Tested functioning on both org and MSDN accounts.


magrande (approved):
Approved !

Issue #225, opened 2021-09-10
Deployment failed when deploying the cluster


lastcoolnameleft:
I'm walking through https://github.com/mspnp/aks-secure-baseline/blob/main/05-aks-cluster.md and on step `Deploy the cluster ARM template.` for `Option 1`, I'm getting the following error:

```
➜  aks-secure-baseline git:(main) RESOURCEID_VNET_CLUSTERSPOKE=$(az deployment group show -g rg-enterprise-networking-spokes -n spoke-BU0001A0008 --query properties.outputs.clusterVnetResourceId.value -o tsv)
                                                                                                                                                                                                                                   [1.1s]
➜  aks-secure-baseline git:(main) az deployment group create -g rg-bu0001a0008 -f cluster-stamp.json -p targetVnetResourceId=${RESOURCEID_VNET_CLUSTERSPOKE} clusterAdminAadGroupObjectId=${AADOBJECTID_GROUP_CLUSTERADMIN_AKS_BASELINE} a0008NamespaceReaderAadGroupObjectId=${AADOBJECTID_GROUP_A0008_READER_AKS_BASELINE} k8sControlPlaneAuthorizationTenantId=${TENANTID_K8SRBAC_AKS_BASELINE} appGatewayListenerCertificate=${APP_GATEWAY_LISTENER_CERTIFICATE_AKS_BASELINE} aksIngressControllerCertificate=${AKS_INGRESS_CONTROLLER_CERTIFICATE_BASE64_AKS_BASELINE} domainName=${DOMAIN_NAME}

(DeploymentFailed) At least one resource deployment operation failed. Please list deployment operations for details. Please see https://aka.ms/DeployOperations for usage details.                                               [7m9.6s]
```

I went to the Portal to see what the specific error was and for the RG, I see:

```
{
  "code": "DeploymentFailed",
  "message": "At least one resource deployment operation failed. Please list deployment operations for details. Please see https://aka.ms/DeployOperations for usage details.",
  "details": [
    {
      "code": "Conflict",
      "message": "Can't update the alert rule right now because there's a similar request in progress on this alert rule. Wait a while and try again. Activity ID: ac83f402-87b0-4c61-86c1-7fb10efbf3ab."
    },
    {
      "code": "Conflict",
      "message": "Can't update the alert rule right now because there's a similar request in progress on this alert rule. Wait a while and try again. Activity ID: 1da932e4-f39b-4c14-89a7-43144ab844f3."
    },
    {
[...]

PR #226, opened 2021-09-10
Change approach to getting AKS_CLUSTER_NAME

lastcoolnameleft:
I had an issue where my deployment failed, so getting the deployment values didn't work.  See https://github.com/mspnp/aks-secure-baseline/issues/225

Instead, I recommend getting the name of the first AKS cluster in the RG as proposed.  This still works even if the deployment fails on a non-critical resource.

### DIFF
--- a/06-gitops.md
+++ b/06-gitops.md
@@ -21,7 +21,7 @@ GitOps allows a team to author Kubernetes manifest files, persist them in their 
 1. Get the cluster name.
 
    ```bash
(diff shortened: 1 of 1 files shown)


ckittel (approved):
Totally okay with this approach.  We were trying to use outputs (because we really shouldn't move on if there was a failure).  We went back and forth on how best to approach this.  Let's run with it like this for a while and see if anyone has objections :)

Thanks!

PR #235, opened 2021-09-21
Add missing preview feature registration and instructions

ckittel:
One of the previews were not listed (I had thought it went GA, so it was removed in a prior commit).  Also had a suggestion to simply list the instructions (in addition to linking to them).


ckittel:
Self merging

PR #236, opened 2021-09-21
Add system CriticalAddonsOnly taint back

ckittel:
This taint was removed from the system node pool before because not all the add-ons supported it.  They all do now, so adding the taint back.


ckittel:
self merging

PR #237, opened 2021-09-21
Add Subject Alternate Name (SAN) extension to the browser-facing cert

ckittel:
The self-signed browser cert didn't have it's SAN populated, so even if you brought the cert into your trusted root store, the cert would still present as invalid.  Also added key usage extensions.

Updated the domain configuration to be saved in the env variables, as it is used across the session.


ckittel:
self merging

PR #239, opened 2021-09-27
Fix save env variables script to work multiple times

ckittel:
The script to save env works only the first time.

For example:
1. Create `*_AKS_BASLINE` env variables
1. Call `saveenv.sh`
1. _reset session_
1. `source aks_baseline.env`
1. `saveenv.sh`

That above results in an _empty_ `aks_baseline.env` file, because the sourced ENVs are not `export`ed.  This fixes that.

### DIFF
--- a/saveenv.sh
+++ b/saveenv.sh
@@ -7,7 +7,7 @@
 cat > aks_baseline.env << EOF
 #!/bin/bash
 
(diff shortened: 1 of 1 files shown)


ckittel:
@alfredoihernandez -- can you review this for me by chance.  Thanks.

alfredoihernandez:
> @alfredoihernandez -- can you review this for me by chance. Thanks.

Looks good, thanks for the fix!

Issue #232, opened 2021-09-18
Key vault with RBAC instead of access policies


stephaneey:
Hi,

Wouldn't it be better to switch to key vault RBAC instead of vault access policies? In the template, you currently create vault policies for the user-assigned identities, where it'd be better to leverage RBAC instead, in order to keep a single pane of glasses to manage authorizations.

Thanks

ckittel:
Indeed @stephaneey, Azure RBAC would have also been a valid solution, and I think it's time we make the migration.  I'll add this to our backlog, and will keep this issue open until we get time to address it.

ckittel:
@stephaneey -- the reference implementation was updated now to use Azure RBAC instead of access policies.  That opened a bit of a can of worms :) but I'm glad it was done.  Thanks for the nudge!

stephaneey:
Well done

DreamRivulet:
It seems that appgw with managed identity  can't work with Key Vault RBAC.

ckittel:
> It seems that appgw with managed identity  can't work with Key Vault RBAC.

Can you add some more context about your experience?

PR #241, opened 2021-09-28
Migrate to Azure RBAC for Key Vault data plane interactions

ckittel:
* Move from Access Policies to Azure RBAC for data plane interactions (Fixes: #232)
* Minor standardization tweak on usage of `export`
* Tightened up Azure Key Vault firewall a bit
* Renamed the app gateway listener SSL cert to be something a bit more descriptive in Key Vault
* Updated some `dependsOn` to include necessary role assignments
* A few RM API version updates

### DIFF
--- a/cluster-stamp.json
+++ b/cluster-stamp.json
@@ -215,7 +215,6 @@
                         }*/
                     ]
                 },
-                "publicNetworkAccess": "Enabled", // Production readiness change: This is only needed because we upload a certificate after deployment as part of the instructions. This can be set to "Disabled" for runtime.
                 "enableRbacAuthorization": true,
                 "enabledForDeployment": false,
                 "enabledForDiskEncryption": false,
--- a/cluster-stamp.json
+++ b/cluster-stamp.json
@@ -201,7 +201,7 @@
                 },
                 "tenantId": "[subscription().tenantId]",
                 "networkAcls": {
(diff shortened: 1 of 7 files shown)


cluster-stamp.json
```diff
@@ -289,14 +262,59 @@
...
+                    ],
+                    "comments": "Grant the Azure Application Gateway managed identity with key vault reader role permissions; this allows pulling frontend and backend certificates.",
+                    "properties": {
+                        "roleDefinitionId": "[variables('keyVaultSecretsUserRole')]",
+                        "principalId": "[reference(resourceId('Microsoft.ManagedIdentity/userAssignedIdentities', 'mi-appgateway-frontend')).principalId]",
+                        "principalType": "ServicePrincipal"
+                    }
+                },
+                {
+                    "type": "providers/roleAssignments",
```
magrande, line 278: Looks like "2020-10-01-preview" is available

magrande (approved):
All good ! Just a single suggestion to upgrade API version.

cluster-stamp.json
ckittel: Yeah, I thought about going through and updating them all in all files, but I knew that was a bit out of scope -- as other ones could be updated as well and this isn't a freshness pass.  The only one I did update was one to help "stage" those firewall changes a bit better.  I'll add this to the backlog, since the new API version also has some additional params

Issue #250, opened 2021-10-28
Azure Policy prevents Traefik Ingress Controller Installation 


kendallroden:
When you attempt to install the Traefik Ingress Controller using the instructions in "08-secret-management-and-ingress-controller.md", the Traefik pods will never come up. After digging, I realized it is because the Traefik deployment is trying to use privilege escalation, which is blocked by Azure Policy. Maybe I am missing something but here is the output:
![Screen Shot 2021-10-28 at 3 12 29 PM](https://user-images.githubusercontent.com/22015135/139328528-b8c40f13-0dde-463f-a725-48d4e704ae4a.png)

ckittel:
Thanks @kendallroden for the suggestion of adding `allowPrivilegeEscalation: false` to help avoid policy-related blocks.  @magrande confirmed upstream that it's a supported configuration, without adding any additional capacities (since this implementation doesn't open any privileged ports).  He added it as part of #252 

Going to close out the issue, but I did see your other ACR notice, where you offered a PR for that -- so we can track that via the PR if you happen to get a chance to open it.

Cheers!

PR #252, opened 2021-11-02
AKS baseline - Kubernetes 1.22 upgrade

magrande:


workload/traefik.yaml
```diff
@@ -297,3 +276,32 @@ spec:
           emptyDir: {}
       nodeSelector:
        agentpool: npuser01
+---
+apiVersion: v1
+kind: List
+metadata:
+  name: traefik-ingress-service
+items:
```
ckittel, line 284: Any reason we moved to a list here instead of just deploying this service like it was before?  Just asking for my own curiosity as I don't remember seeing us do that before anywhere (like in multi-region)

ckittel (changes requested):
Just one question (not blocking though), plus want to make sure that the feedback in #250 is accounted for in this PR (confirmed and added, or not advised and related reference to that provided on the issue thread), which I didn't seen.

ckittel (approved):
lgtm, thanks!

Issue #259, opened 2021-11-15
Flux CrashLoop and Cannot execute Kubectl Logs


scaswell-hirez:
**Bug:**
Flux fails to deploy properly

**Expected Result:** 
Flux launches correctly and deploys the remainder of /cluster-manifests as described

**Actual Result:**
When deploying flux as described in https://github.com/mspnp/aks-secure-baseline/blob/main/06-gitops.md it crashes on startup and enters a CrashLoopBackOff. 

There is no indication as to why this is happening and any attempt to retrieve logs from the pod with kubectl logs results in:

```Error from server (InternalError): Internal error occurred: Authorization error (user=masterclient, verb=get, resource=nodes, subresource=proxy)```

The same error occurs when trying to retrieve logs from the healthy memcached pod.

This is the view of the error from the live logs screen in the portal.
![image](https://user-images.githubusercontent.com/91915937/141867895-05d5763f-1ad0-4074-a49d-ca8f2f148816.png)

Since flux is responsible for deploying everything else in /cluster-manifests this is something of a hard stop regarding the published baseline directions.

**Reproduction:** 
Follow the steps to cleanly deploy the aks-baseline as written.  The flux deployment fails every time.

**Comment by ckittel** [comment]
Can you get the logs for flux from Log Analytics directly and let me know what the crash backoff is?

I don't know what's happening with the Internal Error with your ProxyCall.  But you should still be able to access the logs via direct container log query in Log Analytics.  That'll help you triage the flux deployment crash.  We have not seen that one before, so will be interesting to know what situation you're in that might be causing that.

**Comment by scaswell-hirez** [comment]
So far no.  The only logs available in Log Analytics are from gatekeeper.  I can only assume this is because flux is crashing before even emitting a log message.


Here is my Log Analytics query
```
let startTimestamp = ago(24h);
KubePodInventory
| where TimeGenerated > startTimestamp
| project ContainerID, PodName=Name
| distinct ContainerID, PodName
| join
(
    ContainerLog
    | where TimeGenerated > startTimestamp
)
on ContainerID
// at this point before the next pipe, columns from both tables are available to be "projected". Due to both 
// tables having a "Name" column, we assign an alias as PodName to one column which we actually want
| project TimeGenerated, PodName, LogEntry, LogEntrySource
| order by TimeGenerated desc
| where PodName contains "flux"
```

I'm open to any suggestions on how to diagnose what's happening with flux.

scaswell-hirez:
[...]

PR #261, opened 2021-11-16
Clarify that the nodepools are subject to the authorized IP address ranges as well

ckittel:
As suggested by @scaswell-hirez in #259, add more information about the need to ensure that the public ip addresses that API server traffic is flowing from is allowed in the authorized IP address range.


ckittel:
Documentation, self-merging.

PR #263, opened 2021-11-22
Remind users that the inner-loop scripts are not always up to date

ckittel:


ckittel:
Self merging, wording updates only

PR #269, opened 2021-12-02
Clarify content naming

ckittel:
The legacy naming of "secure baseline" still existed in a few places.


ckittel:
Self merging, naming updates

PR #272, opened 2022-01-04
Use categoryGroup for AFW logs

ckittel:
Updating the API version on the FW to avoid the `categoryGroup` conflict that might happen on redeployment.


magrande (approved):
LGTM !

PR #273, opened 2022-01-04
Update to latest fluxConfiguration RP API version

ckittel:
`httpsCAFile` was renamed to `httpsCACert` in the new API version.  New API version also supports S3 buckets.


magrande (approved):
LGTM !

Issue #275, opened 2022-01-18
Bootstrapping Flux with Private Azure Repos


thepaulmacca:
I'm trying to bootstrap with a private Azure Repo as opposed to a public GitHub repo, and I assume this is a configuration that a lot of others will be using

Any ideas on how I can do that? I've had a look at the [Flux Configurations REST API spec](https://docs.microsoft.com/en-us/rest/api/kubernetesconfiguration/2021-11-01-preview/flux-configurations/create-or-update) and there's not much info on it at the moment. Alot of the current docs are focusing on bootstrapping with the Azure CLI, but nothing else

Thanks

ckittel:
Sure thing.  The features we have exposed on the `fluxConfiguration` extension resource in ARM are listed https://docs.microsoft.com/azure/templates/microsoft.kubernetesconfiguration/fluxconfigurations?tabs=bicep#gitrepositorydefinition (which match your API docs, no surprise there of course).  That said, I'm not actually sure what the right combination of `httpsCACert` and/or `httpsUser` and/or any of the other values there is for private access.  Like you said, the examples all use `k8s-configuration` which completely bypasses the resource provider anyhow and manages the in cluster CRDs directly.

So, I've asked the question internally here to see if I can get some clarity.  If I don't get an answer, I'll go a layer deeper spelunking.

ckittel:
@thepaulmacca -- already have an answer

For SSH+git...

```json
"properties": {
   "gitRepository": {
      "url": "git@github.com:org/repo.git"
   },
   "configurationProtectedSettings" {
      "sshPrivateKey": "base64-encoded-pem-private-key"
   }
}
```

for HTTPS+git

```json
"properties": {
   "gitRepository": {
      "url": "https://github.com/org/repo.git"
   },
   "httpsUser": "<username>"
   "configurationProtectedSettings" {
      "httpsKey": "<PAT token>"
   }
}
```

I'm going to add that to the repo's README tomorrow, at least until we get our docs up to date with this.  Thanks!

thepaulmacca:
Thanks for this @ckittel much appreciated! I've already raised a docs issue but there's not been any movement on it:

https://github.com/MicrosoftDocs/azure-docs/issues/85178

PR #277, opened 2022-02-01
Add private repo boot strapping tip

ckittel:
HT to @thepaulmacca in #275 for the suggestion to include this information.  Product docs will be updated soon with this information, but until then, we'll pull in the highlights here.


ckittel:
Adding additional text only, self-merging.

PR #278, opened 2022-02-01
Adjust firewall rule

ckittel:
Had a NSG rule regression slip through the testing.  Fixing it up.


ckittel:
Self merging, just tested.

PR #279, opened 2022-02-02
Mention the requirement to Base64 Encode the PAT

ckittel:
HT to @ngbrown who called out that the instructions for Personal Access Tokens did not mention the requirement to Base64 encode the value (unlike the SSH instructions).  Adding that bit into the snippits.


ckittel:
Self-merging.

PR #280, opened 2022-02-04
Change nodepoolSubnetResourceIds output to array

thepaulmacca:
I was getting a validation error when deploying a regional hub - as the `nodepoolSubnetResourceIds` parameter in the `hub-regionA.bicep` file is expecting an array, but was getting a string

### DIFF
--- a/networking/spoke-BU0001A0008.bicep
+++ b/networking/spoke-BU0001A0008.bicep
@@ -356,5 +356,7 @@ resource pipPrimaryClusterIp_diagnosticSetting 'Microsoft.Insights/diagnosticSet
 /*** OUTPUTS ***/
 
 output clusterVnetResourceId string = vnetSpoke.id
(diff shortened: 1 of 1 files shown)


ckittel:
Hum, the output is actually designed to be a string because the input to the next template simply gets formatted as an array.  Outputting as an array prevents the correct formatting for the next step.  This is the same way it was when it was in ARM JSON instead.  (Only the `['` and `']` we had in the instructions ... and I just moved them into the bicep to clean it up a bit).  Maybe there is something else going on?

ckittel:
Oh, I see.  I accidently left the command as `-o json` instead of `-o tsv`! If this actually works, I'll be glad to leave it like you have it proposed here. I had it that way originally and it was outputting with newlines in the array and causing some bad command line execution, but I'll test again on Monday.  Maybe I had something else going on as well!  Thanks.

PR #281, opened 2022-02-07
Convert subnet output to array, not an array-like string

ckittel:
Pick up with @thepaulmacca started in #280 and convert that output to an array.


ckittel:
@thepaulmacca -- you were indeed correct that if we do `-o json` like this, it works. I'm not sure why it didn't work the first time for me when I was testing it.  Maybe something else was going on.  Would you happen to be set up to test this specific branch?  If not, I can get someone else on the team to validate.  I tested myself and it was working fine, but would love a second validation.

thepaulmacca:
Sorry @ckittel I've only just caught up on this

I'm not set up on this specific branch, but when I did a deployment earlier today on my own repo, I had to use `-o json` still for validation to pass

I'm based in the UK, so I can validate with this branch in the morning if you like, unless you want it done before, in that case it's probably better if there's anyone around today who can validate it for you

ckittel:
Morning is just fine :)  Enjoy your evening.  I tested it as it's currently in `main` and didn't have any issues -- so I'm not in a rush or anything :)

thepaulmacca:
I tested with this branch earlier using `-o json` and it worked fine for me aswell

Over to you :)

ckittel:
@magrande just did an end-to-end test with this as well, all good from his side.  Merging now.  Thanks again @thepaulmacca!

PR #282, opened 2022-02-08
Gateway Subnet CIDR typo updated to correct range.

skazure:
Typo correction for Gateway Subnet CIDR. update from 10.240.4.0/28 to 10.240.4.16/28


ckittel:
Good catch!  Thanks.

ckittel (approved):

PR #287, opened 2022-02-23
AKS Baseline - Added note about --sdk-auth deprecated parameter and changed Azure login version to v1 in the workflow

magrande:


ckittel (approved):
🚢  LGTM

PR #288, opened 2022-02-24
Fixed fluxV2 extension resource template

teilmeier:
The "configurationSettings" and "configurationProtectedSettings" sections of "Microsoft.KubernetesConfiguration/flux" were defined in the wrong area. Corrected the sections to comply with the schema. https://docs.microsoft.com/en-us/azure/templates/microsoft.kubernetesconfiguration/extensions?tabs=json#template-format
This also fixes the issue of configurations not appearing in the AKS cluster's "GitOps" view in the Azure Portal.


teilmeier:
@ckittel please confirm that this works

ckittel:
Thanks @teilmeier for testing this out and getting the portal experience working again as well!

Issue #293, opened 2022-03-14
App gateway subnet too small?


Justin-DynamicD:
Reviewing this reference architecture, the bicep file to create the subnets in the [spoke network](https://github.com/mspnp/aks-baseline/blob/main/networking/spoke-BU0001A0008.bicep) create a /28 subnet for the app gateway, but [recommended best practices](https://docs.microsoft.com/en-us/azure/architecture/networking/guide/waf-application-gateway#define-application-gateway-subnet-size) say the minimum size should be a /26 and recommend a /24.

Does it make sense to update the deployed example to match said best practices, or is there an overriding recommendation in the scope of AKS?

ferantivero:
thanks for raising this issue @Justin-DynamicD,

it seems we are running a bit small based on the current recommendation, and we should consider right-sizing the appgw subnet based on [this change](https://github.com/MicrosoftDocs/azure-docs/commit/a281aa000620ffce406a219e0549ed5ce6b93abd#diff-d21f682b08540bf8893f9efa18bb2c9c271fda0c5c052239bc4387aa81d7c3a0) and updating our [topology details doc for this](https://github.com/mspnp/aks-baseline-multi-region/blob/main/networking/topology.md).

ckittel:
I've added this as a ticket to our backlog for all of our AKS baseline implementations.  Thanks @Justin-DynamicD for bringing this to our attention and @ferantivero for following up.

Justin-DynamicD:
I appreciate the feedback!

I'm actually going through an exercise of creating a terraform module equivalent of this work (multiple modules actually) and decided this repo was a great way to validate my own research/work.  This was just one area where I found a discrepancy so figured I'd bubble it up and see if maybe I misunderstood something.

Thank you again for answering my questions!

rscottwatson:
@Justin-DynamicD creating a terraform version of this has been on my todo list for a long time.  Are you making your repo for that public?

ckittel:
This has been updated in the AKS baseline for regulated mspnp/aks-baseline-regulated#60.  We still have it on our backlog to apply to the AKS baseline here.  Usually, we do things like this in the AKS baseline first and then port to the "spoke" stories, but we already had some work scheduled on the other, so it made sense from work optimization perspective :)  Keeping this issue open until resolved here as well.

ckittel:
[...]

PR #304, opened 2022-04-19
patch (network topology): [appgw] right size appgw snet

ferantivero:
closes: #293, #577305

## WHY
it seems we are running a bit small based on the current recommendation, and we should consider right-sizing the appgw subnet based on [this change](https://github.com/MicrosoftDocs/azure-docs/commit/a281aa000620ffce406a219e0549ed5ce6b93abd#diff-d21f682b08540bf8893f9efa18bb2c9c271fda0c5c052239bc4387aa81d7c3a0) and updating our [topology details doc for this](https://github.com/mspnp/aks-baseline/blob/main/networking/topology.md).

## WHAT Changed?

- docs
- right size app gw snet from `/28` to `/24`
- change app gw cidr
- whitelisted the new address prefix from in-cluster ingress

## HOW to Test?

we deploy the cluster and the everything works exactly as before. But the app gw subnet now has 251 seats available for app gw instances.


ckittel:
Thanks.  Merging.

PR #305, opened 2022-04-28
Minor fixes & usability improvements

ulkeba:
- Some regions (e.g., also East US 2) require udp/1194 for master to node communication.
- spoke-BU0001A0008.bicep needs to return the IP rather than the Public IP resource ID.
- As minimal user feedback, echo variables to console after setting.

### DIFF
--- a/04-networking.md
+++ b/04-networking.md
@@ -92,8 +92,6 @@ The following two resource groups will be created and populated with networking 
 
    > :book: Now that their regional hub has its first spoke, the hub can no longer run off of the generic hub template. The networking team creates a named hub template (e.g. `hub-eastus2.bicep`) to forever represent this specific hub and the features this specific hub needs in order to support its spokes' requirements. As new spokes are attached and new requirements arise for the regional hub, they will be added to this template file.
 
-   > :warning: If your cluster will be deployed in a region which has not yet konnectivity rolled out for communication from worker to master nodes, you might need to umcomment the firewall rule `pod-to-api-server_udp-1194` in `hub-regionA.bicep` before deployment.
-
    ```bash
    RESOURCEID_SUBNET_NODEPOOLS=$(az deployment group show -g rg-enterprise-networking-spokes -n spoke-BU0001A0008 --query properties.outputs.nodepoolSubnetResourceIds.value -o json)
    echo RESOURCEID_VNET_HUB: $RESOURCEID_SUBNET_NODEPOOLS
--- a/04-networking.md
+++ b/04-networking.md
@@ -76,6 +76,7 @@ The following two resource groups will be created and populated with networking 
 
    ```bash
    RESOURCEID_VNET_HUB=$(az deployment group show -g rg-enterprise-networking-hubs -n hub-default --query properties.outputs.hubVnetId.value -o tsv)
+   echo RESOURCEID_VNET_HUB: $RESOURCEID_VNET_HUB
 
    # [This takes about four minutes to run.]
    az deployment group create -g rg-enterprise-networking-spokes -f networking/spoke-BU0001A0008.bicep -p location=eastus2 hubVnetResourceId="${RESOURCEID_VNET_HUB}"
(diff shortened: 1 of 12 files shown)


networking/hub-regionA.bicep
```diff
@@ -482,6 +482,26 @@ resource fwPolicy 'Microsoft.Network/firewallPolicies@2021-05-01' = {
                 '443'
               ]
             }
+            {
+              ruleType: 'NetworkRule'
+              name: 'pod-to-api-server_udp-1194'
+              description: 'This allows pods to communicate with the API server. Ensure your API server\'s allowed IP ranges support all of this firewall\'s public IPs.'
```
ckittel, line 488:
```suggestion
              description: 'This allows pods to communicate with the API server. Only needed if your cluster isn't using konnectivity.'
```

networking/hub-regionA.bicep
```diff
@@ -482,6 +482,26 @@ resource fwPolicy 'Microsoft.Network/firewallPolicies@2021-05-01' = {
...
+              ]
+              destinationAddresses: [
+                'AzureCloud.${location}' // Ideally you'd list your AKS server endpoints in appliction rules, instead of this wide-ranged rule
+              ]
+              destinationIpGroups: []
+              destinationFqdns: []
+              destinationPorts: [
+                '1194'
+              ]
+            }
```
ckittel, line 504: Technically this is only needed for clusters that are not yet being created with konnectivty, which should be very few at this point.  We have hesitated to add it, because all the clusters that have konnectivity on them now would be getting added outbound firewall allowances that are not necessary. 

What would you say if you had this in here, but commented out with a comment that says "If you cluster does not deploy konnectivity, then this rule will also be needed."  That way for the handful of people that are still getting those clusters, they at least have a solution, and we're not putting a added firewall hole for a majority of people that are not needing the firewall hole?  I know it's not a great solution, either way. :(  Thoughts?

networking/spoke-BU0001A0008.bicep
```diff
@@ -434,4 +434,4 @@ output clusterVnetResourceId string = vnetSpoke.id
 output nodepoolSubnetResourceIds array = [
   vnetSpoke::snetClusterNodes.id
 ]
-output appGwPublicIpAddress string = pipPrimaryClusterIp.id
+output appGwPublicIpAddress string = pipPrimaryClusterIp.properties.ipAddress
```
ckittel, line 437: Yikes, good catch!

ckittel (changes requested):
Thanks for the contribution! 

We had originally avoided the `echo` statements, but we've had multiple requests for them -- so I'm all for trying them out and seeing if we get any feedback one way or the other.  Thanks.

Other feedback below.

networking/hub-regionA.bicep
ckittel: Comment added, marking resolved.

04-networking.md
```diff
@@ -91,8 +92,11 @@ The following two resource groups will be created and populated with networking
 
    > :book: Now that their regional hub has its first spoke, the hub can no longer run off of the generic hub template. The networking team creates a named hub template (e.g. `hub-eastus2.bicep`) to forever represent this specific hub and the features this specific hub needs in order to support its spokes' requirements. As new spokes are attached and new requirements arise for the regional hub, they will be added to this template file.
 
+   > :warning: If your cluster will be deployed in a region which has not yet konnectivity rolled out for communication from worker to master nodes, you might need to umcomment the firewall rule `pod-to-api-server_udp-1194` in `hub-regionA.bicep` before deployment.
```
ckittel, line 95: My only problem with this is that the user isn't going to know at this point, and this will cause anxiety (especially calling this a WARNING -- eek, I'd be nervous if I was going through this the first time -- in a panic because I have no way of checking). I think it might make sense to move this note to one of the "try it" pages after the cluster is deployed. Because you should be able to get through all the bicep deployments without error regardless of this.

First place you really have a hint that it might not be right would be on https://github.com/mspnp/aks-baseline/blob/main/07-bootstrap-validation.md. Can you find a spot there instead? That way if the reader gets through the instructions there, they will be fine, but if they get hung up, you'll have a hint for them.

Also, it doesn't seem to be JUST regional -- I deploy to eastus2 mostly, and haven't had this problem in about three years months. -- So maybe soften up the "regional" aspect of it?

ckittel (changes requested):

ckittel (approved):
LGTM!  💪

PR #306, opened 2022-04-29
Fixup a situation where Azure RBAC role ID was mis-identified

ckittel:
Prevents a deployment failure for AKS resource RBAC role assignments when using Azure RBAC for cluster access.

Also fixes typos in a couple descriptions.

Fixes: #310

### DIFF
--- a/cluster-stamp.bicep
+++ b/cluster-stamp.bicep
@@ -52,7 +52,7 @@ param domainName string = 'contoso.com'
 @minLength(9)
 param gitOpsBootstrappingRepoHttpsUrl string = 'https://github.com/mspnp/aks-baseline'
 
-@description('You cluster will be bootstrapped from this branch in the identifed git repo.')
+@description('You cluster will be bootstrapped from this branch in the identified git repo.')
 @minLength(1)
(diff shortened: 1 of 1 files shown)


cluster-stamp.bicep
```diff
@@ -63,9 +63,6 @@ var acrPullRole = '${subscription().id}/providers/Microsoft.Authorization/roleDe
 var managedIdentityOperatorRole = '${subscription().id}/providers/Microsoft.Authorization/roleDefinitions/f1a07417-d97a-45cb-824c-7a7467783830'
```
ferantivero, line 63: any desire to replicate the same design for this one and others?

ferantivero (approved):
thanks for bug fixing this @ckittel LGTM 🚀 

only one comment to be considered.

cluster-stamp.bicep
ckittel: Not in this pass, as these didn't seem to have problems (still don't really know why) - and I'm keeping the changes scope to just those that I know had the issue.  But yes, I think it would make sense to see if we can do this with the rest of them.

PR #311, opened 2022-05-02
RBAC fix, doc consistency, minor improvements.

ulkeba:


05-bootstrap-prep.md
```diff
@@ -79,7 +79,8 @@ We'll be bootstrapping this cluster with the Flux GitOps agent as installed as a
    Now commit changes to repository.
 
    ```bash
-   git commit -a -m "Update image source to use my ACR instance instead of a public container registry."
+   git add ./cluster-manifests/cluster-baseline-settings/kured.yaml
+   git commit -m "Update image source to use my ACR instance instead of a public container registry."
```
ckittel, line 83: Why make this two commands?  Their shouldn't have been any other changes that aren't already git ignored.  Seems like a maint problem to have to always remember to maintain this changed file list.  We've avoided that elsewhere before.

06-aks-cluster.md
```diff
@@ -9,8 +9,11 @@ Now that your [ACR instance is deployed and ready to support cluster bootstrappi
    > If you cloned this repo, then the value will be the original mspnp GitHub organization's repo, which will mean that your cluster will be bootstraped using public container images. If instead you forked this repo, then the GitOps repo will be your own repo, and your cluster will be bootstrapped using container images references based on the values in your repo's manifest files. On the prior instruction page you had the oppertunity to update those manifests to use your ACR instance. For guidance on using a private bootstrapping repo, see [Private bootstrapping repository](./cluster-manifests/README.md#private-bootstrapping-repository).
 
    ```bash
-   GITOPS_REPOURL=$(git config --get remote.origin.url)
-   echo GITOPS_REPOURL: $GITOPS_REPOURL
+   GITOPS_REPOURL_AKS_BASELINE=$(git config --get remote.origin.url)
+   echo GITOPS_REPOURL_AKS_BASELINE: $GITOPS_REPOURL_AKS_BASELINE
+
+   GITOPS_CURRENT_BRANCH_NAME_AKS_BASELINE=$(git branch --show-current)
+   echo GITOPS_CURRENT_BRANCH_NAME_AKS_BASELINE: $GITOPS_CURRENT_BRANCH_NAME_AKS_BASELINE
```
ckittel, line 16: Why end them in `_AKS_BASLINE` if they are not going to be `export`ed?

10-workload.md
```diff
@@ -46,10 +46,9 @@ The cluster now has an [Traefik configured with a TLS certificate](./08-secret-m
    > You should expect a `403` HTTP response from your ingress controller if you attempt to connect to it _without_ going through the App Gateway. Likewise, if any workload other than the ingress controller attempts to reach the workload, the traffic will be denied via network policies.
 
    ```bash
-   kubectl run curl -n a0008 -i --tty --rm --image=mcr.microsoft.com/azure-cli --overrides='[{"op":"add","path":"/spec/containers/0/resources","value":{"limits":{"cpu":"200m","memory":"128Mi"}}}]' --override-type json
+   kubectl run curl -n a0008 -i --tty --rm --image=mcr.microsoft.com/azure-cli --overrides='[{"op":"add","path":"/spec/containers/0/resources","value":{"limits":{"cpu":"200m","memory":"128Mi"}}}]' --override-type json  --env="DOMAIN_NAME=${DOMAIN_NAME_AKS_BASELINE}"
    
    # From within the open shell now running on a container inside your cluster
-   DOMAIN_NAME="contoso.com" # <-- Change to your custom domain value if a different one was used
    curl -kI https://bu0001a0008-00.aks-ingress.$DOMAIN_NAME -w '%{remote_ip}\n'
```
ckittel, line 52: Slick. love it.  Thanks!

11-validation.md
```diff
@@ -26,8 +26,12 @@ This section will help you to validate the workload is exposed correctly and res
 
 1. Browse to the site (e.g. <https://bicycle.contoso.com>).
 
+   > :bulb: Remember to include the protocol prefix `https://` in the URL you type in the address bar of your browser.
+
    > :bulb: A TLS warning will be present due to using a self-signed certificate. You can ignore it or import the self-signed cert (`appgw.pfx`) to your user's trusted root store.
```
ckittel, line 31: Just to keep it simple.

```suggestion
   > :bulb: Remember to include the protocol prefix `https://` in the URL you type in the address bar of your browser. A TLS warning will be present due to using a self-signed certificate. You can ignore it or import the self-signed cert (`appgw.pfx`) to your user's trusted root store.
```

11-validation.md
```diff
@@ -26,8 +26,12 @@ This section will help you to validate the workload is exposed correctly and res
...
 
+   Refresh the web page a couple of times and observe the value `Host name` displayed at the bottom of the page. As the Traefik Ingress Controller balances the requests between the two pods hosting the web page, the host name will change from one pod name to the other throughtout your queries.
```
ckittel, line 33: Nice touch, thanks!

cluster-stamp.bicep
```diff
@@ -63,9 +63,9 @@ var acrPullRole = '${subscription().id}/providers/Microsoft.Authorization/roleDe
 var managedIdentityOperatorRole = '${subscription().id}/providers/Microsoft.Authorization/roleDefinitions/f1a07417-d97a-45cb-824c-7a7467783830'
 var keyVaultReader = '${subscription().id}/providers/Microsoft.Authorization/roleDefinitions/21090545-7ca7-4776-b22c-e363652d74d2'
 var keyVaultSecretsUserRole = '${subscription().id}/providers/Microsoft.Authorization/roleDefinitions/4633458b-17de-408a-b874-0445c86b69e6'
-var clusterAdminRole = '${subscription().subscriptionId}/providers/Microsoft.Authorization/roleDefinitions/b1ff04bb-8a4e-4dc4-8eb5-8693973ce19b'
-var serviceClusterUserRole = '${subscription().subscriptionId}/providers/Microsoft.Authorization/roleDefinitions/4abbcc35-e782-43d8-92c5-2d3f1bd2253f'
-var clusterReaderRole = '${subscription().subscriptionId}/providers/Microsoft.Authorization/roleDefinitions/7f6c6a51-bcf8-42ba-9220-52d62157d7db'
+var clusterAdminRoleId = 'b1ff04bb-8a4e-4dc4-8eb5-8693973ce19b'
+var serviceClusterUserRoleId = '4abbcc35-e782-43d8-92c5-2d3f1bd2253f'
+var clusterReaderRoleId = '7f6c6a51-bcf8-42ba-9220-52d62157d7db'
```
ckittel, line 68: If you can revert the change to this file, the merge conflict will be resolved and the problem will also be resolved :)  The pattern of using `existing` is what we'll likely want to do for the rest of them as well, but for now having it mixed like this will be fine.

You'll want to keep the `empty()` check change you made though for the namespace RBAC condition.

saveenv.sh
```diff
@@ -4,10 +4,12 @@
 # the page they are created on. Then a user can source this file to restore those environment 
 # variables if their shell session is reset for some reason.
 
-cat > aks_baseline.env << EOF
+DIR_NAME=$(dirname "$0")
+
+cat > $DIR_NAME/aks_baseline.env << EOF
```
ckittel, line 9: Just curious what problem this was solving.  We've used this `saveenv.sh` script here and in the CSE hackathons -- never ran into an issue.  Any pointers as to what benefit this version has?  We'll want to replicate it in the other projects that use this once I know. :)

ckittel (changes requested):
Sweet, nice set of quality of life improvements!  Thanks!  Please see the suggestions here.

06-aks-cluster.md
ulkeba: That's true. I have lost the `export` when copying from another "one shot script" where I had it. Will add `export`.

ckittel: are you sure it's needed?  I didn't catch where you ended up needing it later.

saveenv.sh
ulkeba: I was in a subdirectory and called it using the relative path `../saveenv.sh` which caused the `aks_baseline.env` files to exist twice -- one (older) in the project root directory, the other one in the subdirectory from which I called `savenenv.sh` using the relative path. With this fix, the location of `aks_baseline.env` should always be well-defined and avoid irritation.

ckittel: Ah, that makes sense.  Following the directions would have never gotten someone into that state.  But I can see that if someone was "off the rails" a bit for whatever reason (good/bad/otherwise), this would help.  Cool.

cluster-stamp.bicep
ulkeba: Your solution is much nicer! :-)
Will revert it.

05-bootstrap-prep.md
ulkeba: I had changed SKUs and availability zones (you may know why...) in cluster-stamp.bicep that I had not wanted to commit. But I do not insist on that, if you think it may be better to leave it as it was.

ckittel: Understood the reasoning in your case.  Yeah, let's just leave it at one command to simplify the steps (and maint).  If someone is "off the path" for other reasons, they can figure out how to just commit what they needed to commit.  We'll stay true on the happy path :)

06-aks-cluster.md
ulkeba: You are right, it will only be needed in the following step if you strictly follow the process.

But I did a couple of deployment (especially while I was adding the name prefixes and chasing this role lookup bug) and did that step over and over again. Since I had multiple terminals open, I switched context by running `saveenv.sh` (which is lovely -- never seen that before!) and sourcing the `aks_baseline.env` to restore the context. 

Later, another deployment failed because the branch variable that I had added was not passed correctly and so, the bootstrapping tried to get scripts from the main branch that had not the replaced repo FQDNs.

This is why I thought (just for consistency) it might make sense to also put it into the environment.

ckittel: Gotcha.  Yeah, our policy so far has been that we only export the variable if it's going to be used on a "later page" in the steps.  The pages are pretty natural 'stopping points' for someone to go wander away so higher likelihood of a console blip (Looking at you, codespaces!).  So that was the rule we put into place, never on ones where the command is used just on the same page (usually as the very next step).

ckittel: Thanks for the feedback on the `saveenv.sh` -- that was born out of codespaces dropping the console frequently.  The Commercial Software Engineering crew came up with this for the codespaces-based workshops they run with the AKS baseline.  It's very handy.

ulkeba: So you prefer to have export and suffix removed?
(As far as I see, these are the only two variables without export and suffix...)

ulkeba: Oh now -- there are some more.

ulkeba: Hmmm... Well, I love the idea of not needing to think about whether a variable will be written to the env file or not. But this is a personal preference and I'll be fine with whatever you decide.

ckittel: Correct,  the only variables that should be `export`'d are ones that will be accessed on a future page.

Examples of ones that are not exported can be found on
- the AAD page (only used on that page)
- The networking page (only used on that page)
- This GITOPS url (and now branch) 
- The Bootstrap validation page 
- etc

Bascially only those that have `_AKS_BASELINE` should have `export` and the only variables that should have `_AKS_BASELINE` are those that are used on a future page.

ckittel: This is consistent with the walkthroughs on the AKS Baseline for multi-region and AKS Baseline for PCI-DSS 3.2.1 (regulated).  Any chance in this policy should probably be a backlog item and addressed across them all.  But for now this pattern has been met with success.

ulkeba: Understood. 👍

05-bootstrap-prep.md
ckittel: Thanks for reverting this.

06-aks-cluster.md
ckittel: @ulkeba -- in your revert of this, you missed a spot that you had changed lower on the page.

cluster-stamp.bicep
ckittel: Thanks, revert looks good.  Kept the empty check.

06-aks-cluster.md
ulkeba: Good catch.
Sorry.

ckittel (approved):
Thanks again, time to :shipit:.

06-aks-cluster.md
ckittel: Resolved!

PR #312, opened 2022-05-02
Remove Inner Loop files

ckittel:
These files always lag behind and are getting more and more out of sync, especially with the changes to bicep now.  Just going to delete them.


ckittel:
Self-merging.

Issue #314, opened 2022-05-04
cluster-stamp.bicep produces some warnings on deployment


akulich:
Here are the warnings produced by `cluster-stamp.bicep` deployment (line numbers are slightly different as it was modified with couple of additional parameters):
```
cluster-stamp.bicep(178,18) : Warning BCP081: Resource type "Microsoft.ContainerService/managedClusters/namespaces@2022-01-02-preview" does not have types available.
cluster-stamp.bicep(217,5) : Warning BCP037: The property "etag" is not allowed on objects of type "SavedSearchProperties". Permissible properties include "functionAlias", "functionParameters", "tags". If this is an inaccuracy in the documentation, please report it to the Bicep Team. [https://aka.ms/bicep-type-issues]
cluster-stamp.bicep(230,5) : Warning BCP037: The property "etag" is not allowed on objects of type "SavedSearchProperties". Permissible properties include "functionAlias", "functionParameters", "tags". If this is an inaccuracy in the documentation, please report it to the Bicep Team. [https://aka.ms/bicep-type-issues]
cluster-stamp.bicep(907,5) : Warning BCP073: The property "scope" is read-only. Expressions cannot be assigned to read-only properties. If this is an inaccuracy in the documentation, please report it to the Bicep Team. [https://aka.ms/bicep-type-issues]
cluster-stamp.bicep(932,5) : Warning BCP073: The property "scope" is read-only. Expressions cannot be assigned to read-only properties. If this is an inaccuracy in the documentation, please report it to the Bicep Team. [https://aka.ms/bicep-type-issues]
cluster-stamp.bicep(951,5) : Warning BCP073: The property "scope" is read-only. Expressions cannot be assigned to read-only properties. If this is an inaccuracy in the documentation, please report it to the Bicep Team. [https://aka.ms/bicep-type-issues]
cluster-stamp.bicep(970,5) : Warning BCP073: The property "scope" is read-only. Expressions cannot be assigned to read-only properties. If this is an inaccuracy in the documentation, please report it to the Bicep Team. [https://aka.ms/bicep-type-issues]
cluster-stamp.bicep(994,5) : Warning BCP073: The property "scope" is read-only. Expressions cannot be assigned to read-only properties. If this is an inaccuracy in the documentation, please report it to the Bicep Team. [https://aka.ms/bicep-type-issues]
[...]

PR #317, opened 2022-05-19
Azure Policy Refresh

ckittel:
Refreshing our Azure Policy assignments.  This PR will not address adding custom in-cluster policies.

Changes:

- No longer building tenant resource IDs for policy definitions by hand, now using built-in `tenantResourceId()` function
- Added Gatekeeper constraint names to bicep file for easy cross referencing
- included `flux-system` as it's coming from an extension in more `excludedNamespaces` on policy assignments
- went from `audit` -> `Audit` and `deny` -> `Deny` to align with the defaults better.
- Documented all known policy violations
- Updated API versions related to Azure Policy assignments
- Populated description on all Azure Policy assignments
- Ensure `take()` is applied to all policy names to prevent a long RG name breaking the assignment name
- Updated the following previously applied Azure Policy for Kubernetes policy assignments:
   - `K8sAzureContainerAllowedImages`
      - Added proper RegEx escaping for `.` in the allow list
      - Removed `azurearcfork8s.azurecr.io/azurearcflux/images/stable/.+$` as those flux images are now sourced from MCR.
   - `K8sAzureContainerLimits`
      - No longer including `cluster-baseline-settings`
      - Updated limits to match cluster workload requirements.
   - `K8sAzureReadOnlyRootFilesystem`
      - Moved from `Audit` to `Deny`
      - Added specific exceptions for the cluster's workloads
- Assigned the following built-in Azure Policy for Kubernetes policies:
   - `K8sAzureHostFilesystem` (`Deny`)
   - `K8sAzureExternalIPs` (`Deny`)
   - `K8sAzureBlockEndpointEditDefaultRole` (`Audit`)
   - K8sAzureBlockDefault (`Audit`)
- Assigned the following built-in Azure Policy policies:
   - Authorized IP ranges should be defined on Kubernetes Services
   - Kubernetes Services should be upgraded to a non-vulnerable Kubernetes version
   - Role-Based Access Control (RBAC) should be used on Kubernetes Services
   - Azure Kubernetes Service Clusters should use managed identities
   - Container registries should have anonymous authentication disabled
   - Container registries should have local admin account disabled

Unrelated changes bundled in this PR:

- variable reduction in bicep template when references would have worked.
- fixed all bicep warnings (fixes: #314)
- Removed some empty `dependsOn: []`

### DIFF
--- a/acr-stamp.bicep
+++ b/acr-stamp.bicep
@@ -72,16 +72,6 @@ resource spokeVirtualNetwork 'Microsoft.Network/virtualNetworks@2021-05-01' exis
   }
 }
 
-// Built-in policy: Container registries should have anonymous authentication disabled.
-var pdAnonymousContainerRegistryAccessDisallowedId = tenantResourceId('Microsoft.Authorization/policyDefinitions', '9f2dea28-e834-476c-99c5-3507b4728395')
-
-// Built-in policy: Container registries should have repository scoped access token disabled.
-var pdAccessTokenContainerRegistryAccessDisallowedId = tenantResourceId('Microsoft.Authorization/policyDefinitions', 'ff05e24e-195c-447e-b322-5e90c9f9f366')
-
-// Built-in policy: Container registries should have local admin account disabled.
-var pdAdminAccountContainerRegistryAccessDisallowedId = tenantResourceId('Microsoft.Authorization/policyDefinitions', 'dc921057-6b28-4fbe-9b83-f7bec05db6c2')
-
-
 /*** RESOURCES ***/
 
 // This Log Analytics workspace will be the log sink for all resources in the cluster resource group. This includes ACR, the AKS cluster, Key Vault, etc. It also is the Container Insights log sink for the AKS cluster.
@@ -96,7 +86,8 @@ resource laAks 'Microsoft.OperationalInsights/workspaces@2021-06-01' = {
   }
 }
 
-// Ensure the container registry does not allow anon access (Azure RBAC only is allowed)
+// Apply the built-in 'Container registries should have anonymous authentication disabled' policy. Azure RBAC only is allowed.
+var pdAnonymousContainerRegistryAccessDisallowedId = tenantResourceId('Microsoft.Authorization/policyDefinitions', '9f2dea28-e834-476c-99c5-3507b4728395')
 resource paAnonymousContainerRegistryAccessDisallowed 'Microsoft.Authorization/policyAssignments@2021-06-01' = {
   name: guid(resourceGroup().id, pdAnonymousContainerRegistryAccessDisallowedId)
   location: 'global'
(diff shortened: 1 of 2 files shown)


cluster-stamp.bicep
```diff
@@ -1214,47 +1214,6 @@ resource paDisallowNamespaceUsage 'Microsoft.Authorization/policyAssignments@202
   }
 }
 
-// Built-in policy: AKS clusters should gate deployment of vulnerable images' policy at the resource group level.
```
ferantivero: what about keeping them in with a conditional? Not asking for a change just giving an alternative since it was originally intended to be included.

cluster-stamp.bicep
```diff
@@ -957,7 +957,7 @@ resource paEnforceHttpsIngress 'Microsoft.Authorization/policyAssignments@2021-0
     policyDefinitionId: pdEnforceHttpsIngressId
     parameters: {
       excludedNamespaces: {
-        value: [] // None
```
ferantivero: is the `value` field required to be declarative in here to work properly? If not, any difference between this and empty `dependson`? Not asking for change trying to understand the rationale/thought-process.

ferantivero (commented):
w00t nicely done @ckittel, a few comments below to be discussed.

cluster-stamp.bicep
```diff
@@ -196,7 +196,6 @@ resource alaRgRecommendations 'Microsoft.Insights/activityLogAlerts@2020-10-01'
     enabled: true
     description: 'All azure advisor alerts'
   }
-  dependsOn: []
```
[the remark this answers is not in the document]
ckittel: Yeah, that's not a resource property (resources don't have these properties), that's a ARM json/bicep feature.

cluster-stamp.bicep
```diff
@@ -1051,14 +1075,13 @@ resource paEnforceImageSource 'Microsoft.Authorization/policyAssignments@2021-06
     parameters: {
       allowedContainerImagesRegex: {
         // If all images are pull into your ARC instance as described in these instructions you can remove the docker.io entries.
-        value: '${acr.name}\\.azurecr\\.io/.+$|mcr\\.microsoft\\.com/.+$|azurearcfork8s\\.azurecr\\.io/azurearcflux/images/stable/.+$|docker\\.io/weaveworks/kured.+$|docker\\.io/library/.+$'
```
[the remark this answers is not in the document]
ckittel: It does pull the image, but now from `mcr` instead of their own ACR instance.  They are updated via helm charts that are pushed as part of their extension.

cluster-stamp.bicep
ckittel: We have some work to do to include the new Defender feature more completely.  Right now, these policies error out on evaluation when the feature isn't fully enabled -- including pricings in security center (which we don't configure here -- but do in the regulated).  Since we really aren't telling a defender story in the baseline, I'm only including the bits that don't cause Azure Policy failures to surface.  If we come back with bringing Defender into the baseline in a more "official" way, then this policy will need to come back.  This policy only really works if you've pushed all images through the new ACR pipeline audit feature -- which we very much don't do here.  Adding it as an optional param is probably a bit overkill at this point, since we don't set the user up for success with it.

cluster-stamp.bicep
ckittel: yeah, it's important here to have `[]` in this case.  This is two fold.

1) This helps be clear that we are NOT excluding any namespaces (including `kube-system`).  The intent here is to show that we want this rule cluster-wide.  Practically speaking, `gatekeeper-system` and `kube-system` are hardcoded in the `Deployment` as being skipped anyway for gatekeeper, but the default policy position is to list them explicitly anyway just in case that changes on the AKS-managed `Deployment` -- but in this case, this policy is safe to apply to all the places, so making it clear (intentional).

2) and most importantly, the policy definition's default is the common trio of namespaces, so if we don't include this `excludedNamespaces` it comes in as `null` and then gets the common trio added, which is not what we want in this case.

cluster-stamp.bicep
ferantivero: thanks for sharing the context and we can definitely get rid of it for now as this not adding value in the end.

acr-stamp.bicep
```diff
@@ -114,25 +105,8 @@ resource paAnonymousContainerRegistryAccessDisallowed 'Microsoft.Authorization/p
   }
 }
 
-// Ensure the container registry does not allow sas token access (Azure RBAC only is allowed)
-resource paAccessTokenContainerRegistryAccessDisallowed 'Microsoft.Authorization/policyAssignments@2021-06-01' = {
```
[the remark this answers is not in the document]
ferantivero: Thanks for sharing the details and yeah, it seems the preview is not ready for us I guess.

Im resolving this now.

cluster-stamp.bicep
ckittel: Sounds good. Thanks.

cluster-stamp.bicep
```diff
@@ -886,167 +883,481 @@ resource sqrPodFailed 'Microsoft.Insights/scheduledQueryRules@2018-04-16' = {
...
   }
-  dependsOn: []
 }
 
-// Applying the 'AKS Linux Restrictive' policy to the resource group
+// Resource Group Azure Policy Assignments - Azure Policy for Kubernetes Policies
+
+// Applying the built-in 'Kubernetes cluster pod security restricted standards for Linux-based workloads' initiative at the resource group level.
+// Constraint Names: K8sAzureAllowedSeccomp, K8sAzureAllowedCapabilities, K8sAzureContainerNoPrivilege, K8sAzureHostNetworkingPorts, K8sAzureVolumeTypes, K8sAzureBlockHostNamespaceV2, K8sAzureAllowedUsersGroups, K8sAzureContainerNoPrivilegeEscalation
+var psdAKSLinuxRestrictiveId = tenantResourceId('Microsoft.Authorization/policySetDefinitions', '42b8ef37-b724-4e24-bbc8-7a7708edfe00')
```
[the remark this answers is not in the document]
ckittel: Okay, moved back up in https://github.com/mspnp/aks-baseline/pull/317/commits/dc5e3d530ded4ed2ef32f9e56be2a87cdaa62238. LMK what you think.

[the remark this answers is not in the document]
ferantivero: awesome, please let me resolve this one.

PR #319, opened 2022-05-23
Document opinion on sending kube-scheduler to Log Analytics

ulkeba:
* Add `kube-scheduler` to the bicep template -- default `false`
* Add a note about when it might be good to set it to `true`


cluster-stamp.bicep
```diff
@@ -1567,6 +1567,10 @@ resource mc_diagnosticSettings  'Microsoft.Insights/diagnosticSettings@2021-05-0
         category: 'guard'
         enabled: true
       }
+      {
+        category: 'kube-scheduler'
+        enabled: true
```
ckittel, line 1572:
```suggestion
        enabled: false  // Only enable while tuning or triaging issues with working scheduling. On a normally operating cluster there is minimal value, relative to the log capture cost, to always keeping this enabled.
```

ckittel (changes requested):
We reviewed this with the K8s Champs/SME group, and they overwhelmingly suggested that this not be enabled by default on clusters.  It should be enabled while someone is just starting up their cluster for the first time, during the "figure it out" phase, triage, etc -- but not as a standard runtime setting.  The concern putting it in the baseline here is that it will be implied that it's a standard "on" value instead of "on, when needed."  We updated the AKS baseline RA on Friday to call this out specifically now, for kube-scheduler.

The cost (and noise) is just too much for a normally operating cluster for these logs.

06-aks-cluster.md
```diff
@@ -19,6 +19,8 @@ Now that your [ACR instance is deployed and ready to support cluster bootstrappi
...
 
+   :bulb: The cluster will send [diagnostic information](https://docs.microsoft.com/en-us/azure/aks/monitor-aks) of categories _cluster-autoscaler_, _kube-controller-manager_, _kube-audit-admin_ and _guard_ to the Log Analytics Workspace deployed as part of the [bootstrapping step](./05-bootstrap-prep.md). If you plan to inspect the behavior of the scheduler, you might want to also enable the log category _kube-scheduler_ in `cluster-stamp.bicep`. :exclamation: Note that this category is quite verbose and might impact the cost of your workspace.
```
ckittel, line 22: Can this get moved to another section?  I know "catching them before deploying" is what you were hoping to do.  But  I don't want a non-security related item here to be distracting from the deployment steps (which is what they are here to follow).  Putting this note in or around the [Azure Monitor Insights & Logs](https://github.com/mspnp/aks-baseline/blob/main/11-validation.md#validate-cluster-azure-monitor-insights-and-logs ) section would be better. 

```suggestion
   :bulb: The cluster will send [diagnostic information](https://docs.microsoft.com/azure/aks/monitor-aks) of categories _cluster-autoscaler_, _kube-controller-manager_, _kube-audit-admin_ and _guard_ to the Log Analytics Workspace deployed as part of the [bootstrapping step](./05-bootstrap-prep.md). If you need to inspect the behavior of the Kubernetes scheduler, enable the log category _kube-scheduler_ in `cluster-stamp.bicep`. This category is quite verbose and will impact the cost of your workspace.
```

ckittel (changes requested):

ckittel (approved):
LGTM. Once again, thanks for the contribution!

PR #321, opened 2022-05-24
Curl image wasn't obeying the new policy restrictions

ckittel:
Update the run command to include read only root filesystem to avoid being blocked by policy.


ckittel:
Self-merging.

PR #323, opened 2022-05-25
Update SecretProviderClass API version

ckittel:
* Updates the inline `SecretProviderClass` to use the GA API version
* Adds explicit `false` for `useVMManagedIdentity`


ferantivero (approved):
w00t nicely done @ckittel LGTM :shipit:

PR #337, opened 2022-08-10
Baseline Updates

v-fearam:
1- I read about the AKS new version 

    https://azure.microsoft.com/en-us/updates/public-preview-kubernetes-release-124-support-in-aks/
    https://kubernetes.io/blog/2022/04/07/upcoming-changes-in-kubernetes-1-24/

    Some points are not impacting the project, others i'm not sure (but I don't think so). 
    After changing and test it works.
2- _Note_ :  A double "see" verb was detected and fixed on the documentation.
3-  kured upgraded, https://github.com/weaveworks/kured


ckittel (approved):
LGTM!  Thanks for the updates.

PR #345, opened 2022-09-15
Deploy custom policy as part of cluster deployment

ulkeba:
PR contains:
- Motivation to deploy a custom policy (Ingress hosts must have a well-defined domain suffix.)
- Code to deploy the custom policy using Azure Policy.
- Validation procedure, illustrating how creation of a violating resource will be rejected.


ulkeba:
I decided for a Base64-coded specification of the Gatekeeper policy. The advantage is that the bicep is self-containing and we do not have a dependency of an external file we'd need to download from a GitHub repository. The disadvantage is that users cannot really read it. 

We could (a) leave it like that, (b) commit the policy into the repository (for reference only) and link in the explanation, (c) replace the encoding with the URL to the (readable) spec. 

Any thoughts on this?

nested_K8sCustomIngressTlsHostsHaveDefinedDomainSuffix.bicep
```diff
@@ -0,0 +1,159 @@
+targetScope = 'subscription'
+
+/*** PARAMETERS ***/
+
+@description('The identifying name this custom policy is deployed with.')
+@minLength(1)
+param policyName string = 'K8sCustomIngressTlsHostsHaveDefinedDomainSuffix'
```
ckittel, line 7: Why make this a param with a default?  You don't use the default, and you provide this exact value, and the file is named is this, and the base64 is encoded text basically uses this as well.  I'd just simplify and hard code it on line 13 so this isn't a one-off.

nested_K8sCustomIngressTlsHostsHaveDefinedDomainSuffix.bicep
```diff
@@ -0,0 +1,159 @@
...
+
+@description('The identifying name this custom policy is deployed with.')
+@minLength(1)
+param policyName string = 'K8sCustomIngressTlsHostsHaveDefinedDomainSuffix' 
+
+/*** RESOURCES ***/
+
+resource pdK8sCustomIngressTlsHostsHaveDefinedDomainSuffix 'Microsoft.Authorization/policyDefinitions@2021-06-01' = {
+  scope: subscription()
+  name: policyName
```
ckittel, line 13:
```suggestion
  name: 'K8sCustomIngressTlsHostsHaveDefinedDomainSuffix'
```

cluster-stamp.bicep
```diff
@@ -1411,6 +1413,44 @@ resource paManagedIdentitiesEnabled 'Microsoft.Authorization/policyAssignments@2
   }
 }
 
+// Deploying and applying the custon policy 'Kubernetes cluster ingress TLS hosts must have defined domain suffix' as defined in 'custom-policy_K8sCustomIngressTlsHostsHaveDefinedDomainSuffix.bicep'
```
ckittel, line 1416:
```suggestion
// Deploying and applying the custom policy 'Kubernetes cluster ingress TLS hosts must have defined domain suffix' as defined in nested_K8sCustomIngressTlsHostsHaveDefinedDomainSuffix.bicep
```

cluster-stamp.bicep
```diff
@@ -1411,6 +1413,44 @@ resource paManagedIdentitiesEnabled 'Microsoft.Authorization/policyAssignments@2
...
+resource paK8sIngressTlsHostsHaveSpecificDomainSuffix 'Microsoft.Authorization/policyAssignments@2021-06-01' = {
+  dependsOn: [
+    modK8sIngressTlsHostsHaveDefinedDomainSuffix
+  ]
+  name: guid(policyNameIngressTlsHostsSuffix, resourceGroup().id, clusterName)
+  location: 'global'
+  scope: resourceGroup()
+  properties: {
+    displayName: take('[${clusterName}] ${modK8sIngressTlsHostsHaveDefinedDomainSuffix.outputs.poliyName}', 120)
+    description: modK8sIngressTlsHostsHaveDefinedDomainSuffix.outputs.poliyDescription
```
ckittel, line 1436:
```suggestion
    description: modK8sIngressTlsHostsHaveDefinedDomainSuffix.outputs.policyDescription
```

cluster-stamp.bicep
```diff
@@ -1411,6 +1413,44 @@ resource paManagedIdentitiesEnabled 'Microsoft.Authorization/policyAssignments@2
...
+    policyDefinitionId: modK8sIngressTlsHostsHaveDefinedDomainSuffix.outputs.poliyId
```
ckittel, line 1437: Do I need to lone you a `c` key?  :)

```suggestion
    policyDefinitionId: modK8sIngressTlsHostsHaveDefinedDomainSuffix.outputs.policyId
```

cluster-stamp.bicep
```diff
@@ -1828,6 +1869,7 @@ resource mc 'Microsoft.ContainerService/managedClusters@2022-03-02-preview' = {
     paOldKuberentesDisabled
     paRbacEnabled
     paManagedIdentitiesEnabled
+
```
ckittel, line 1872:
```suggestion
```

nested_K8sCustomIngressTlsHostsHaveDefinedDomainSuffix.bicep
```diff
@@ -0,0 +1,159 @@
...
+        }
+      }
+    }
+  }
+}
+
+
+output poliyId string = pdK8sCustomIngressTlsHostsHaveDefinedDomainSuffix.id
+output poliyName string = pdK8sCustomIngressTlsHostsHaveDefinedDomainSuffix.properties.displayName
+output poliyDescription string = pdK8sCustomIngressTlsHostsHaveDefinedDomainSuffix.properties.description
```
ckittel, line 159:
```suggestion
output policyId string = pdK8sCustomIngressTlsHostsHaveDefinedDomainSuffix.id
output policyName string = pdK8sCustomIngressTlsHostsHaveDefinedDomainSuffix.properties.displayName
output policyDescription string = pdK8sCustomIngressTlsHostsHaveDefinedDomainSuffix.properties.description
```

11-validation.md
```diff
@@ -40,6 +40,45 @@ If instead Kubernetes RBAC is backed directly by Azure AD, then you'll need to e
 
 No matter which backing store you use, the user assigned to the group will then be able to `az aks get-credentials` to the cluster and you can validate that user is limited to a _read only_ view of the a0008 namespace.
 
+## Validate Azure Policy
+
+Built-in as well as custom policies are applied to the cluster as part of the [cluster deployment step](./06-aks-cluster.md) to ensure that workloads deployed to the cluster comply with the team's governance rules. Policy assignments with effect [`audit`](https://docs.microsoft.com/en-us/azure/governance/policy/concepts/effects#audit) will create a warning in the activity log and show violations in the Azure Policy blade in the portal, providing an aggregated view of the compliance state and the option to identify violating resources. Policy assignments with effect [`deny`](https://docs.microsoft.com/en-us/azure/governance/policy/concepts/effects#deny) will be enforced with the help of [Gatekeeper's admission controller webhook](https://open-policy-agent.github.io/gatekeeper/website/docs/) by denying API requests that would violate a policy otherwise.
```
ckittel, line 45:
```suggestion
Built-in as well as custom policies are applied to the cluster as part of the [cluster deployment step](./06-aks-cluster.md) to ensure that workloads deployed to the cluster comply with the team's governance rules. Policy assignments with effect [`audit`](https://learn.microsoft.com/azure/governance/policy/concepts/effects#audit) will create a warning in the activity log and show violations in the Azure Policy blade in the portal, providing an aggregated view of the compliance state and the option to identify violating resources. Policy assignments with effect [`deny`](https://learn.microsoft.com/azure/governance/policy/concepts/effects#deny) will be enforced with the help of [Gatekeeper's admission controller webhook](https://open-policy-agent.github.io/gatekeeper/website/docs/) by denying API requests that would violate a policy otherwise.
```

11-validation.md
```diff
@@ -40,6 +40,45 @@ If instead Kubernetes RBAC is backed directly by Azure AD, then you'll need to e
...
+
+### Steps
+
+1. Try to add a second `Ingress` resource to your workload namespace with the following command (note, that the host specified in the `rules` and the `tls` sections defines a domain name with suffix `invalid-domain.com` rather than the domain suffix you defined for your setup when you [created your certificates](./02-ca-certificates.md)).
```
ckittel, line 49:
```suggestion
1. Try to add a second `Ingress` resource to your workload namespace with the following command.

    Notice that the host value specified in the `rules` and the `tls` sections defines a domain name with suffix `invalid-domain.com` rather than the domain suffix you defined for your setup when you [created your certificates](./02-ca-certificates.md)).
```

11-validation.md
```diff
@@ -40,6 +40,45 @@ If instead Kubernetes RBAC is backed directly by Azure AD, then you'll need to e
...
+           pathType: Prefix
+           backend:
+             service:
+               name: aspnetapp-service
+               port:
+                 number: 80
+   EOF
+   ```
+
+2. Inspect the error message and remark that Gatekeepers admission webhook rejects `bu0001a0008-00.aks-ingress.invalid-domain.com` as incompliant host.
```
ckittel, line 76:
```suggestion
2. Inspect the error message and remark that Gatekeeper's admission webhook rejects `bu0001a0008-00.aks-ingress.invalid-domain.com` as incompliant host.
```

11-validation.md
```diff
@@ -40,6 +40,45 @@ If instead Kubernetes RBAC is backed directly by Azure AD, then you'll need to e
...
+
+   ```bash
```
ckittel, line 78:
~~~suggestion
   ```output
~~~

cluster-stamp.bicep
```diff
@@ -1411,6 +1413,44 @@ resource paManagedIdentitiesEnabled 'Microsoft.Authorization/policyAssignments@2
...
 
+// Deploying and applying the custon policy 'Kubernetes cluster ingress TLS hosts must have defined domain suffix' as defined in 'custom-policy_K8sCustomIngressTlsHostsHaveDefinedDomainSuffix.bicep'
+// Note: Policy definition must be deployed as module since policy definitions require a targetScope of 'subscription'.
+
+module modK8sIngressTlsHostsHaveDefinedDomainSuffix 'nested_K8sCustomIngressTlsHostsHaveDefinedDomainSuffix.bicep' = {
+  name: 'modK8sIngressTlsHostsHaveDefinedDomainSuffix'
+  scope: subscription()
+  params: {
+    policyName: policyNameIngressTlsHostsSuffix
+  }
```
ckittel, line 1424: To go with the suggestion in the module.

```suggestion
```

cluster-stamp.bicep
```diff
@@ -66,6 +66,8 @@ var aksIngressDomainName = 'aks-ingress.${domainName}'
 var aksBackendDomainName = 'bu0001a0008-00.${aksIngressDomainName}'
 var isUsingAzureRBACasKubernetesRBAC = (subscription().tenantId == k8sControlPlaneAuthorizationTenantId)
 
+var policyNameIngressTlsHostsSuffix = 'K8sCustomIngressTlsHostsHaveDefinedDomainSuffix'
+
```
ckittel, line 70: To go with the suggestion in the module.

```suggestion
```

cluster-stamp.bicep
```diff
@@ -1411,6 +1413,44 @@ resource paManagedIdentitiesEnabled 'Microsoft.Authorization/policyAssignments@2
...
+  params: {
+    policyName: policyNameIngressTlsHostsSuffix
+  }
+}
+
+resource paK8sIngressTlsHostsHaveSpecificDomainSuffix 'Microsoft.Authorization/policyAssignments@2021-06-01' = {
+  dependsOn: [
+    modK8sIngressTlsHostsHaveDefinedDomainSuffix
+  ]
+  name: guid(policyNameIngressTlsHostsSuffix, resourceGroup().id, clusterName)
```
ckittel, line 1431: Depends on will already be added based on the existing references to the `outputs`.

```suggestion
  name: guid(modK8sIngressTlsHostsHaveDefinedDomainSuffix.outputs.policyId, resourceGroup().id, clusterName)
```

ckittel (changes requested):
I think this is great, @ulkeba.  THANK YOU.  Can you consider my suggestions below?

nested_K8sCustomIngressTlsHostsHaveDefinedDomainSuffix.bicep
ulkeba: My first attempt was exactly as you suggested: Use the policyId from the nested output as first parameter to the `guid(.)` function in `cluster-stamp.bicep` (this would even be the most consistent setup copared to the other policy assignments that also use the policy ID here).

However, the Bicep compiler rejected the usage of `modK8sIngressTlsHostsHaveDefinedDomainSuffix.outputs` in `name: guid(...)` when calling `az deployment group create -g rg-bu0001a0008 -f cluster-stamp.bicep...` the with following message:

```
Error BCP120: This expression is being used in an assignment to the "name" property of the "Microsoft.Authorization/policyAssignments" type, which requires a value that can be calculated at the start of the deployment. Properties of modK8sIngressTlsHostsHaveDefinedDomainSuffix which can be calculated at the start include "name".
```

So I was looking for another value to the `guid(.)` function which would be a sufficiently unique to this assignment to avoid guid collisions with other resources. The best idea I came up with was to pull the name (that should be fairly unique) out as parameter, define it in `cluster-stamp.bicep` and use as parameter to the module (to avoid divergence of the name in policyAssignment and policyDefinition). (Although I agree that it technically makes no sense to declare a parameter here.)

Just to be sure I did not miss anything, I checked out your suggestion and tried to deploy again but got the same message.

Let me know what you think...

ckittel: Bummer, I was wondering if that was going to happen.  Well, the hard coded policy name in there as the first param should be good enough, That's unfortunate, but not really the end of the world :)

ulkeba: Okay. Sounds good.

ulkeba:
> I decided for a Base64-coded specification of the Gatekeeper policy. The advantage is that the bicep is self-containing and we do not have a dependency of an external file we'd need to download from a GitHub repository. The disadvantage is that users cannot really read it.
> 
> We could (a) leave it like that, (b) commit the policy into the repository (for reference only) and link in the explanation, (c) replace the encoding with the URL to the (readable) spec.
> 
> Any thoughts on this?

Any thoughts on this? ☝️

ckittel:
> Any thoughts on this? ☝️

I have no problem with it base64 encoded like you have it, without having to maintain a shadow copy.  I was going to link to a base64 decoder in a comment, but didn't see any that MSFT owned publicly so I passed on that.  If you wanted to give the user a hint, you can add a comment that says the body is base64 encoded, and if they wish to view it, to open it in a decoder.  But I wouldn't worry about it.

ckittel:
Looks like there are some merge conflicts @ulkeba -- if you can address those, and if you wanted to do any suggestions for the user to check out a decoder for the contents, then we'd be good to go with this.

ckittel (approved):
This is great!  Thank you, @ulkeba!

PR #347, opened 2022-09-16
docs -> learn URL updates

ckittel:


ckittel:
self merging this find and replace

Issue #342, opened 2022-08-24
Update AzureBastionSubnet size


ckittel:
The subnet for Azure Bastion is currently a `/27` when it needs to be a `/26` due to service changes suggesting a larger subnet.  This reference implementation doesn't deploy Bastion (but leaves a hole for it) -- and it would be good if that hole was right-sized.

PR #353, opened 2022-10-07
Fix Azure Bastion Subnet to be /26 (instead of /27)

ckittel:
Fixes: #342

When this reference implementation was created, `/27` was the correct size for Azure Bastion.  The latest recommendation is `/26` -- luckily enough room was left :) 

* Updated the `AzureBastionSubnet` subnet size, and the related IP table
* Minor formatting & link changes on the IP table


networking/topology.md
```diff
@@ -30,15 +30,15 @@ In the future, this VNet might hold more subnets like [ACI Provider instance] su
...
+| Azure Firewall Subnet (AzureFirewallSubnet) | -            | [59]               | -                    | -          | -                        | 100           | 100                 | 0                     | 0                         | 5                                        | 0                   | 64                    | 64                 | 26                 | 10.200.0.0/26  | 10.200.0.0  | 10.200.0.63   |
+| Azure Bastion Subnet (AzureBastionSubnet)   | -            | [50<sup>2</sup>]   | -                    | -          | -                        | 100           | 100                 | 0                     | 0                         | 5                                        | 0                   | 64                    | 64                 | 26                 | 10.200.0.128/26 | 10.200.0.128 | 10.200.0.191  |
```
hallihan, line 41: The superscript here seems to indicate that there is more information below, but # 2 in Additional Considerations is talking about a different subnet.  Not new for this PR, but maybe we can clean that up?

hallihan (approved):
Looks good, noticed one inconsistency that would be good to address before merging.

networking/topology.md
ckittel: The superscript link works, it's literally the whole text, when clicked, goes to that data source, but I see what you mean how it could be confused with the numbered list below.  Let me fix that!  Good idea.

ckittel: Double fix, removed the (now) unnecessary superscript -- they were used as differentiators before, and also converted the numbered list into a bullet list.

ckittel:
Thanks for the review, @hallihan!

PR #355, opened 2022-10-12
Fix retry interval - portal and logs get angry with null

ckittel:
null technically defaults to the sync interval, but portal doesn't like it and the logs throw needless messages without it.  Setting to 300 (5 minutes)


ckittel:
self merging

PR #361, opened 2022-12-07
Fix url typos

ckittel:
Fix one typo and update to latest branch naming on upstream repo to avoid a redirect.


ckittel:
self merging

PR #391, opened 2023-09-20
treafik 2.10

v-fearam:
following recommendation
https://doc.traefik.io/traefik/migration/v2/#kubernetes-crds
traefik.containo.us is deprecated, and its support will end starting with Traefik v3


workload/traefik.yaml
```diff
@@ -44,16 +45,17 @@ rules:
     verbs:
       - update
   - apiGroups:
+      - traefik.io
```
ckittel, line 48: Ah, good catch.

ckittel (approved):

PR #404, opened 2024-03-06
Align resources to resource group location

ckittel:
- By default, use resource group location to drive the location of resources, based on reliability recommendations from resource provider teams.
- This change required breaking out the hubs and spokes into resource-group-per-region aligned hubs (resource group name changes to include region)
- Had to remove the region restrictions since you cannot mix defaults with allow lists

Unrelated changes:

- Updated oldest supported version of `kubectl` (must be within 1 of the cluster version)
- Minor typos fixed

Tested end to end:
image

image

Work Item: https://dev.azure.com/msft-skilling/Content/_workitems/edit/217262


05-bootstrap-prep.md
```diff
@@ -10,15 +10,15 @@ Container registries often have a lifecycle that extends beyond the scope of a s
...
+In addition to Azure Container Registry being deployed to support bootstrapping, this is where any other resources that are considered not tied to the lifecycle of an individual cluster is deployed. Azure Container Registry is one example as talked about above. Another example could be an AKS Backup Vault and backup artifacts storage account which likely would exist prior to and after any individual AKS cluster's existence. When designing your pipelines, ensure to isolate components by their lifecycle watch for singletons in an architecture. These are typically resources like regional logging sinks, supporting global routing infrastructure, and so on. This is in contrast to potentially transiently/replaceable components, like the AKS cluster itself. *This implementation does not represent a complete separation of stamp vs regional resources, but is fairly close. Deviations are strictly for ease of deployment in this walkthrough instead of as examples of guidance.*
```
skabou, line 21:
```suggestion
In addition to Azure Container Registry being deployed to support bootstrapping, this is where any other resources that are considered not tied to the lifecycle of an individual cluster is deployed. Azure Container Registry is one example as talked about above. Another example could be an AKS Backup Vault and backup artifacts storage account which likely would exist prior to and after any individual AKS cluster's existence. When designing your pipelines, ensure to isolate components by their lifecycle watch for singletons in an architecture. These are typically resources like regional logging sinks, supporting global routing infrastructure, and so on. This is in contrast with potentially transient/replaceable components, like the AKS cluster itself. *This implementation does not represent a complete separation of stamp vs regional resources, but is fairly close. Deviations are strictly for ease of deployment in this walkthrough instead of as examples of guidance.*
```

skabou (approved):
Looks good - Very thorough!  Just those couple editing nits

ckittel:
Thanks for the improvements and the review!  Merging now and closing out the task.

PR #411, opened 2024-07-15
Update API versions in Bicep files

johndowns:
This PR updates the majority of the Bicep resources to use the most recent non-preview API version available, with the following exceptions:
- Scheduled query rules: There are some major changes in the API for this resource type, which I will handle in a separate PR.
- AKS cluster: The cluster itself needs to use a preview version (2024-03-02-preview) to enable the `nodeRestriction` feature.

The PR also adds strong typing for the Azure Policy definitions, and simplifies some of the places these definitions are referenced. Finally, it updates a place where the incorrect seed is used for the `guid()` function when creating a role assignment.

I have tested deployment using these updated files.


ckittel (approved):
All changes, including the `clusterControlPlaneIdentityName` changeover looks good. I don't remember why we did what we did with that name vs using the identity's guid, but I can't think of a tenable reason for it, so good to see the change.

PR #412, opened 2024-07-15
Enable zone redundancy on ACR

johndowns:
Now that the Central US region supports availability zones, I believe we can safely make the container registry zone-redundant. This PR removes the previous note that the feature is unavailable.


ckittel (approved):
<3

PR #417, opened 2024-07-17
[WIP] Version updates

johndowns:
This PR updates the versions of AKS (to 1.30) and Traefik (to 3.0).


ckittel (approved):
Glad to see a bunch of the config didn't need to be changed between v2 and v3.  I think when we went from v1 to v2, the configuration section (traefik.toml) had a large overhaul.

Nice!

(Aside, not a blocker, noticed that 3.1 was JUST released, can never keep up :))

PR #418, opened 2024-07-18
Update Bicep API version for scheduled query rules

johndowns:
This PR fixes an outstanding issue with the Bicep definitions for scheduled query rules. The latest API versions require changes to the definition of the resources.


ckittel (approved):
thanks!

Good to 🚢 after the URL update (if you wish to apply)

PR #421, opened 2024-07-30
[WIP] Enable Azure Monitor metrics collection

johndowns:
This PR enables Azure Monitor metrics collection for the cluster, which in turn enables [Network Observability](https://learn.microsoft.com/azure/aks/network-observability-overview).


johndowns:
Pausing work on this PR for now. In order to support Network Observability, the cluster also needs to be associated with Prometheus and Grafana, which requires further discussion and work.

PR #422, opened 2024-08-06
Skip Log Analytics query validation during Bicep deployment

johndowns:
This PR works around a potential deployment race condition.

We deploy a scheduled query rule to detect failed pods. However, the query references a Kusto table that only exists after Container Insights is deployed and running, which takes an unpredictable amount of time.

To work around deployment failures, the Azure Monitor PG advised skipping query validation.


ferantivero (approved):
w00t @johndowns, LGTM 🚢

ferantivero:
#sign-off

PR #424, opened 2024-08-12
Refactor some resources into modules

johndowns:
This PR refactors some of the Bicep code into modules, to reduce the noise in the main Bicep file.

I've intentionally kept this quite light. The changes in this PR are:

**cluster-stamp.bicep**
- Extract the Azure Monitor alerts into their own module.
- Extract the Azure Policy assignments into their own module.
- Rename the existing modules to clarify their purpose.
- Fix some minor typos.


workload-team/cluster-stamp.bicep
```diff
@@ -28,6 +28,7 @@ param clusterAuthorizedIPRanges array = []
 @description('AKS Service, Node Pool, and supporting services (KeyVault, App Gateway, etc) region. This needs to be the same region as the vnet provided in these parameters. This defaults to the resource group\'s location for higher reliability.')
 param location string = resourceGroup().location
 
+@description('The Kubernetes version for the AKS cluster.')
 param kubernetesVersion string = '1.30'
```
ferantivero, line 32: I know this is not part of this PR, but since we added a description to this I was wondering whether we should add allowed values as well or simple convert this into a variable.

Feel free to mark as won't fix if this is not too relevant at this point in time.

workload-team/modules/alerts.bicep
ferantivero, line 1: would you consider changing this to something like `cluster-monitoring` or similar?

Feel free to reject this proposal if you are good with the current naming.

workload-team/modules/policies.bicep
ferantivero, line 1: I would've called this one something like `cluster-governance` or similar. 

Please let me try to backup a bit why I'm proposing changing in the naming. 

I understand we want to separate the policies from the managed cluster by modularizing. As a result this adds readability when looking at the code, and that is a best practice when this is about code maintenance.

But based on my understanding the main guidance is around infrastructure and it is that clusters should not be created without organization governance in first place. 

That's why I think if we could keep tied _"the main files"_ by using some sort of naming convention could be a great common ground between what we're fairly trying to achieve and the previous version in which a single file delivered an important piece of guidance in an implicit but very practical way. 

Consider this renaming just a suggestion for you to pick out or reject based on your preference.

ferantivero (changes requested):
great changes in here @johndowns, it looks really nice. 

Please let me share some feedback on top of this.

docs/aks-baseline_details.drawio.svg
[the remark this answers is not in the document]
johndowns: No, updating the relative file paths.

workload-team/acr-stamp.bicep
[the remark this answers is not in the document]
johndowns: I'll defer to @ckittel, who likely has context on the history of this one.

workload-team/cluster-stamp.bicep
johndowns: I added a description because it's the only param without a description currently, so it was purely for consistency.

Thinking about it, maybe this should be a var instead of a param - we don't necessarily want people changing the version and expecting it to work. By using a var, we indicate that this is the version we've tested with. What do you think?

workload-team/modules/policies.bicep
johndowns: If I was doing this for a real production solution I would probably agree (and with the "monitoring" suggestion too) - I'd also do a lot more modularisation, as we've previously discussed. But in this particular case, the intention is just to remove some of the noise out of the main Bicep file to make it a little easier to follow.

I feel like "governance" is a wider area than just policies, so again I feel like it might be overstating it to name the module that way when it only contains one aspect of governance.

workload-team/cluster-stamp.bicep
ferantivero: I would say yes to a var instead

johndowns: Updated.

ferantivero (approved):
great modularization @johndowns, LGTM 🚢

PR #425, opened 2024-08-15
Update network profile to use Azure CNI Overlay

johndowns:
This PR switches the baseline reference implementation to use [Azure CNI Overlay](https://learn.microsoft.com/azure/aks/azure-cni-overlay), which aligns with current guidance for the majority of customers.


workload-team/cluster-stamp.bicep
```diff
@@ -1786,6 +1786,8 @@ resource mc 'Microsoft.ContainerService/managedClusters@2024-03-02-preview' = {
     enablePodSecurityPolicy: false
     networkProfile: {
       networkPlugin: 'azure'
+      networkPluginMode: 'overlay'
```
ckittel, line 1789: Does this change, in any way, affect the details on <https://github.com/johndowns/aks-baseline/blob/cni-overlay/network-team/topology.md>?

ckittel (changes requested):

workload-team/cluster-stamp.bicep
johndowns: Technically, no - but I think it's a good idea to add some new content to that file to explain what's going on with this extra CIDR range. I've pushed an update.

ckittel (approved):

ferantivero (approved):
nice change in here @johndowns LGTM 🚢  

Please let's consider merging after accommodating address spaces in based on what we recommend from the reference architecture (new texts for CNI overlay) and our offline discussion on top of the cluster configured nodes

PR #428, opened 2024-09-03
feat (deploy): [dns] parametrize ILB private Ipv4

ferantivero:
- add new cluster-stamp param for lib ipv4 with default value
- output assigned dns ipv4 to be later used when creating k8s objs
- modify deployment steps to use the ipv4 cluster-stamp deployment output value

## Test

image

closes: #302538


workload-team/cluster-stamp.bicep
```diff
@@ -39,6 +39,9 @@ param gitOpsBootstrappingRepoHttpsUrl string = 'https://github.com/mspnp/aks-bas
 @minLength(1)
 param gitOpsBootstrappingRepoBranch string = 'main'
 
+@description('The AKS cluster Internal Load Balancer IP Address')
+param clusterInternalLoadBalancerIpAddress string = '10.240.4.4'
```
johndowns, line 43: It looks like this parameter isn't set in the calling script - should it be?

ferantivero: I'm not quite sure whether it should @johndowns  but this can definitely be an alternative

Please let's analyze from the following commit what is the preference. Happy to revert or keep this as is. 

done | addressed from d5910b1145d0775319f2bed5491ba6b45d6c5797

docs/deploy/09-secret-management-and-ingress-controller.md
```diff
@@ -76,6 +76,13 @@ Previously you have configured [workload prerequisites](./08-workload-prerequisi
    az acr import --source docker.io/library/traefik:v3.1 -n $ACR_NAME_AKS_BASELINE
    ```
 
+1. Retrieve the fixed IPv4 address for the AKS internal load balancer ingress controller's service.
```
johndowns, line 79:
```suggestion
1. When the cluster was deployed, the Bicep deployment decided on the fixed private IP address to use for the AKS internal load balancer ingress controller's service. Retrieve that IP address so we can configure it in the Traefik deployment, which will create the load balancer.
```

johndowns (approved):
Just one suggestion around the documentation to help to explain what's going on. Apart from that, LGTM!

docs/deploy/09-secret-management-and-ingress-controller.md
ferantivero: good catch @johndowns, thanks for the feedback.

ferantivero:
#sign-off

PR #433, opened 2024-10-08
feat (deploy): [monitoring] [Prometheus] add rule groups for linux

ferantivero:


workload-team/cluster-stamp.bicep
ferantivero, line 1: check whether current manifest files are double login metrics if this pr is merged.

Additionally, consider whether existent queries should start using this new metrics instead of LA

PR #437, opened 2024-11-29
feat (cluster): [monitoring ci] change ContainerLogV2 to basic plan

ferantivero:
## WHY

we do wanted to experiment with the new `basic` plan that reduce costs when ingesting gibs into log analytics 

## WHAT Change?

- configure `ContainerLogV2` table with basic plan at creation time 

## Test

image

image
image
image

image

image

image

closes: #319587

### DIFF
--- a/workload-team/acr-stamp.bicep
+++ b/workload-team/acr-stamp.bicep
@@ -51,6 +51,17 @@ resource laAks 'Microsoft.OperationalInsights/workspaces@2023-09-01' = {
       dailyQuotaGb: -1 // No daily cap (configure alert below if enabled)
     }
   }
+
+  resource containerLogV2 'tables' = {
+    name: 'ContainerLogV2'
+    properties: {
+      totalRetentionInDays: 30
+      plan: 'Basic'
+      schema: {
+        name: 'ContainerLogV2'
+      }
+    }
+  }
 }
 
 // Add a alert rule if the log analytics workspace daily data cap has been reached.


johndowns:
Hey @ckittel would you please be able to review this one, since you proposed the work initially? Does it cover the scope of what you had in mind? It relates to https://dev.azure.com/msft-skilling/Content/_workitems/edit/319587

ckittel (changes requested):
What's the cost impact on this based on things like Azure Monitor for Containers (Does that use the available logs? I'm not even sure) and our query-based alerts?  We'd probably only do this if the overall cost is cheaper for this scenario.  We'll want to address this question not only here, but as a decision point in the RA.

As far as the implementation goes, this is fine.  But need the "why/benefit/tradeoff" bit understood, since the work item was first and foremost a "evaluate if we should" task.

ferantivero:
> What's the cost impact on this based on things like Azure Monitor for Containers (Does that use the available logs? I'm not even sure) and our query-based alerts? We'd probably only do this if the overall cost is cheaper for this scenario. We'll want to address this question not only here, but as a decision point in the RA.
> 
> As far as the implementation goes, this is fine. But need the "why/benefit/tradeoff" bit understood, since the work item was first and foremost a "evaluate if we should" task.

great questions, please let's conduct this further analysis offline.

ckittel (approved):
LGTM, as long as @johndowns can coordinate updates to the RA based on :

* Impact to Microsoft Defender for Containers -- no impact, as the logs it pulls from it not ContainerLogV2
* Impact to query-based alerts -- no impact in the actual deployment, but if a user wanted to write query-based alerts against ContainerLogV2 they would need to evaluate if Basic is appropriate.
* In most customer situations, the number of writes far outweights the number of reads, and nodes can be very chatty to ContainerLogV2 -- so it's a great place to start for cost optimization efforts and only move away from Basic once the cost is no longer in favor of this model.

ferantivero:
#sign-off

PR #443, opened 2025-08-22
feat (infra): [network] disable default outbound access

ferantivero:
## WHY

we wanted to explicitly disable outbound access, means no default outbound access for all subnet except those delegated or managed. We wanted to ensure that baseline is working enforcing this setup since default access is being retired in September

## WHAT Changed

- `firewall`, `gateway`, `bastion`, `appgw`, `privateendpoints`, `cluster-nodes` and `ingress` subnets are now marked with default outbound access as `false` 

## TEST

image

hub

image

spoke
image


ckittel (approved):
Wonderful, thanks!

PR #450, opened 2025-12-02
feat (cluster): [control plane dns] disable public FQDN and link private DNS zone to hub for Bastion

ferantivero:
## WHY

While revisiting the reference architecture we found that Azure Bastion now supports tunneling to AKS private clusters without requiring the public FQDN. Disabling the public FQDN removes a publicly discoverable DNS name that resolves to the private API server IP, reducing the cluster attack surface.

However, when the public FQDN is disabled, the only resolvable endpoint is the privatelink.*.azmk8s.io FQDN, which requires the private DNS zone to be linked to any VNet that needs to resolve it, including the hub VNet where Bastion resides. This PR adds that link, enabling Bastion to work with the public FQDN disabled.

## WHAT Changed?

- [x] disable public FQDN for the private AKS cluster (`enablePrivateClusterPublicFQDN: false`)
- [x] link the AKS private DNS zone to the hub VNet so Bastion can resolve the private API server endpoint
- [x] add `hubVnetResourceId` parameter to `cluster-stamp.bicep`
- [x] update deploy docs to capture and pass hub VNet resource ID
- [x] update managed cluster RP API version to the latest

## TEST

Tested all four combinations in francecentral to validate the fix:

| # | Public FQDN | Hub DNS Link | Bastion works? |
|---|---|---|---|
| 1 | true | No | :white_check_mark: (public DNS resolves to private IP) |
| 2 | true | Yes | :white_check_mark: (redundant but fine) |
| 3 | false | No | :x: (TLS handshake timeout, hub cannot resolve privatelink FQDN) |
| 4 | false | Yes | :white_check_mark: (the fix in this PR) |

Test case 3 confirms the failure mode: `az aks bastion` opens the tunnel successfully (platform-level operation), but `kubectl` commands fail with TLS handshake timeout because the private FQDN cannot be resolved from the hub VNet.

Test case 4 confirms the fix: adding the hub VNet link to the private DNS zone restores full Bastion connectivity with the public FQDN disabled. This is the configuration shipped in this PR.


workload-team/cluster-stamp.bicep
```diff
@@ -960,9 +989,6 @@ resource mc 'Microsoft.ContainerService/managedClusters@2024-03-02-preview' = {
       azureKeyVaultKms: {
         enabled: false // Not enabled in the this deployment, as it is not used. Enable as needed.
       }
-      nodeRestriction: {
-        enabled: true // https://kubernetes.io/docs/reference/access-authn-authz/admission-controllers/#noderestriction
-      }
       defender: {
```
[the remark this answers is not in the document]
johndowns: I agree with Copilot here - @ferantivero can you please confirm what the intention was here?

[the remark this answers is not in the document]
ferantivero: good catch @johndowns, please lets take into account we are intentionally dropping that from the web app routing  pr

done | addressed ddbdf1066b74cba46895d80b85caed53da4c9c2e

workload-team/cluster-stamp.bicep
```diff
@@ -169,6 +173,20 @@ resource targetVirtualNetwork 'Microsoft.Network/virtualNetworks@2023-11-01' exi
   }
 }
 
+/*** EXISTING HUB RESOURCES ***/
+
+// Hub resource group
+resource hubResourceGroup 'Microsoft.Resources/resourceGroups@2024-03-01' existing = {
```
johndowns:
```suggestion
resource hubResourceGroup 'Microsoft.Resources/resourceGroups@2025-04-01' existing = {
```

workload-team/cluster-stamp.bicep
```diff
@@ -169,6 +173,20 @@ resource targetVirtualNetwork 'Microsoft.Network/virtualNetworks@2023-11-01' exi
...
+  scope: subscription()
+  name: split(hubVnetResourceId, '/')[4]
+}
+
+// Hub virtual network
+resource hubVirtualNetwork 'Microsoft.Network/virtualNetworks@2023-11-01' existing = {
```
johndowns:
```suggestion
resource hubVirtualNetwork 'Microsoft.Network/virtualNetworks@2025-07-01' existing = {
```

johndowns (dismissed):
Thanks - just two small things and then this is good to go.

workload-team/cluster-stamp.bicep
ferantivero: good catch @johndowns, updated 

done | addressed from 9b567b2

workload-team/cluster-stamp.bicep
ferantivero: good catch @johndowns, updated 

done | addressed from 9b567b2

johndowns (approved):

